PR #16, opened 2023-07-09
WIP on Sequencer V2 design

aaclayton:


module/data/action.mjs
```diff
@@ -1026,64 +1038,50 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
+          .play()
+        break;
+      case "single":
+        for ( const outcome of this.outcomes.values() ) {
+          if ( !outcome.rolls.length ) continue;
+          const hit = outcome.rolls.some(r => r.isSuccess || (r.data.damage?.total > 0));
+          const targetToken = CrucibleAction.#getTargetToken(outcome.target);
+          sequence.effect()
+            .preset(id, {source: originToken, target: targetToken, missed: !hit, ...options})
+            .play()
```
Haxxer, line 1076: You may benefit from moving the `.play()` call out of the `for` loop, as each outcome would play the entire assembled Sequence several times, which could duplicate effects that have already been played.

module/visual-effects/sequencer-presets.mjs
```diff
@@ -0,0 +1,54 @@
...
+    const dbEntry = Sequencer.Database.getEntry(spellPath);
+    if ( !dbEntry ) {
+      console.warn(`Expected Sequencer database entry for "${spellPath}" was not found.`);
+      return effect;
+    }
+
+    // Standard Fan effect
+    effect.file(spellPath)
+      .atLocation({x: template.x, y: template.y}, {cacheLocation: true})
+      .rotate(template.direction - 90)
```
Haxxer, line 46: I believe you can use `.atLocation(template, { cacheLocation: true })` and `.rotateTowards(template, { cacheLocation: true })` here - it achieves the same thing. The location based methods such as these are intelligent and know that the `.atLocation()` call is for the start point of cones and rays, and any `.stretchTo()` or similar targets the end point of the template.

PR #19, opened 2024-07-10
Subdivide grid

dev7355608:


aaclayton (approved):
I might change this to only happen automatically IF the current configuration is `(units === "ft") && (distance === 5)` but I can take that over after merge. Thanks!

PR #29, opened 2025-02-24
Change `SceneDimensions#rect` and `SceneDimensions#sceneRect` to `PIXI.Rectangle`

dev7355608:
Same change as in https://github.com/foundryvtt/foundry-vtt/pull/3293.

### DIFF
--- a/module/documents/scene.mjs
+++ b/module/documents/scene.mjs
@@ -34,12 +34,10 @@ export default class CrucibleScene extends Scene {
 
     // Define Scene dimensions
     return {
-      width, height,
-      size: grid.size,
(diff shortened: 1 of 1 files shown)


module/documents/scene.mjs
```diff
@@ -34,12 +34,10 @@ export default class CrucibleScene extends Scene {
 
     // Define Scene dimensions
     return {
-      width, height,
-      size: grid.size,
-      uiSize: sourceGrid.size,
```
dev7355608, line 39: I couldn't find this property being used anywhere so I removed.

aaclayton (approved):

PR #259, opened 2025-07-29
Added settings for extending character creation

danielrab:
The way it currently works is
- The indexes of all the item packs are iterated through to find the packs that contain items appropriate for a given setting.
- The built-in packs are used as defaults for the settings.
- The packs are displayed using their paths in the format `folder -> subfolder -> pack`.
- All the settings are marked as requiring a reload because, at least the talent tree doesn't get updated otherwise.
Please let me know if any changes are needed to the logic, the code style, or anything else.

### DIFF
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -83,6 +83,7 @@ Hooks.once("init", async function() {
               folder = folder.folder;
             }
             potentialPacks[pack.metadata.id] = name;
+            break;
           }
         }
       }
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -70,83 +70,51 @@ Hooks.once("init", async function() {
   /**
    * Configurable properties of the system which affect its behavior.
    */
-  game.settings.register("crucible", "usedAncestries", {
-    name: "SETTINGS.UsedAncestriesName",
-    hint: "SETTINGS.UsedAncestriesHint",
-    scope: "world",
-    config: true,
-    type: new foundry.data.fields.SetField(new foundry.data.fields.StringField({ required: true, choices: () => {
+  function itemPacksByType(type) {
       let potentialPacks = {};
       for (const pack of game.packs) {
         if (pack.metadata.type !== "Item") continue;
         for (const item of pack.index) {
-          if (item.type === "ancestry") {
+          if (item.type === type) {
             potentialPacks[pack.metadata.id] = pack.metadata.id;
           }
         }
       }
       return potentialPacks;
-    }})),
+  }
+  game.settings.register("crucible", "ancestrySources", {
+    name: "SETTINGS.AncestrySourcesName",
+    hint: "SETTINGS.AncestrySourcesHint",
+    scope: "world",
+    config: true,
+    type: new foundry.data.fields.SetField(new foundry.data.fields.StringField({ required: true, choices: () => itemPacksByType("ancestry")})),
     default: [SYSTEM.COMPENDIUM_PACKS.ancestry],
     requiresReload: true,
   });
-  game.settings.register("crucible", "usedBackgrounds", {
-    name: "SETTINGS.UsedBackgroundsName",
-    hint: "SETTINGS.UsedBackgroundsHint",
+  game.settings.register("crucible", "backgroundSources", {
+    name: "SETTINGS.BackgroundSourcesName",
+    hint: "SETTINGS.BackgroundSourcesHint",
     scope: "world",
     config: true,
-    type: new foundry.data.fields.SetField(new foundry.data.fields.StringField({ required: true, choices: () => {
-      let potentialPacks = {};
-      for (const pack of game.packs) {
-        if (pack.metadata.type !== "Item") continue;
-        for (const item of pack.index) {
-          if (item.type === "background") {
-            potentialPacks[pack.metadata.id] = pack.metadata.id;
-          }
-        }
-      }
-      return potentialPacks;
-    }})),
+    type: new foundry.data.fields.SetField(new foundry.data.fields.StringField({ required: true, choices: () => itemPacksByType("background")})),
     default: [SYSTEM.COMPENDIUM_PACKS.background],
     requiresReload: true,
   });
-  game.settings.register("crucible", "usedSpells", {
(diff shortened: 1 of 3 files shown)


crucible.mjs
```diff
@@ -70,16 +70,72 @@ Hooks.once("init", async function() {
...
+    default: [SYSTEM.COMPENDIUM_PACKS.talent],
+    requiresReload: true,
+  });
+  
   crucible.CONFIG = {
     /**
      * Configuration of compendium packs which are used as sources for system workflows.
      * @type {Record<string, Set<string>>}
      */
     packs: {
```
aaclayton, line 134: The `init` hook is too early to retrieve world-level settings. We need to move the definition of `CONFIG.packs` later on into the `"setup"` hook (but before we initialize the talent tree).

crucible.mjs
```diff
@@ -70,16 +70,72 @@ Hooks.once("init", async function() {
   /**
    * Configurable properties of the system which affect its behavior.
    */
+  function itemPacksByType(type) {
```
aaclayton, line 73: One thing we need to be quite cautious of here is not to invalidate the setting value if a module is disabled. Suppose a user adds a custom pack to `ancestrySources` like `"myModule.myPack"` and then `myModule` gets disabled. In that event `"myModule.myPack"` will no longer be a valid choice for the setting and (I'm pretty sure) will cause some problems.

I think you need to test this case carefully and ensure that registering a pack here and then disabling the module that registered it doesn't break the system or cause the entire setting to get reset to default. 

We likely want the setting itself to be tolerant of any string value but present a multi-select element to the user which only includes (a) current values and (b) currently available pack options.

Make sense?

crucible.mjs
```diff
@@ -70,16 +70,72 @@ Hooks.once("init", async function() {
...
+      let potentialPacks = {};
+      for (const pack of game.packs) {
+        if (pack.metadata.type !== "Item") continue;
+        for (const item of pack.index) {
+          if (item.type === type) {
+            let name = pack.metadata.label;
+            let folder = pack.folder;
+            while (folder) {
+              name = `${folder.name} -> ${name}`
```
aaclayton, line 82: I think rather than `folder -> name` I would prefer to have `name (id)`. For example `"Crucible Talents (crucible.talents)"`

lang/en.json
```diff
@@ -823,6 +823,15 @@
 "SETTINGS.AutoConfirmSelf": "Self-Target Actions Only",
 "SETTINGS.AutoConfirmAll": "Auto-Confirm All Actions",
 
+"SETTINGS.AncestrySourcesName": "Ancestry Sources",
+"SETTINGS.AncestrySourcesHint": "Which packs will be used as sources of ancestries during character creation.",
```
aaclayton, line 827: 1. Field labels and hints should canonically use lowercase `label` and `hint` as the final parts of their localization keys. This allows them to (potentially, but not required) work with prelocalization.
2. Let's use a bit of nesting rather than keeping the keys flat. For example:
```js
"SETTINGS": {
  "COMPENDIUM_SOURCES": {
    "ancestry": {
      "label": "Ancestry Sources",
      "hint": "Which packs will be used as sources of ancestries during character creation."
    },
    ...
  }
}
```

crucible.mjs
```diff
@@ -70,16 +70,72 @@ Hooks.once("init", async function() {
...
+              folder = folder.folder;
+            }
+            potentialPacks[pack.metadata.id] = name;
+            break;
+          }
+        }
+      }
+      return potentialPacks;
+  }
+  game.settings.register("crucible", "ancestrySources", {
```
aaclayton, line 92: I think I'm _okay_ with 4 separate settings, I think I would _prefer_ one setting `crucible.compendiumSources` with a data structure like:
```js
crucible.compendiumSources = {
  ancestry: [],
  background: [],
  spell: [],
  talent: []
}
```

crucible.mjs
```diff
@@ -70,16 +70,72 @@ Hooks.once("init", async function() {
...
     /**
      * Configuration of compendium packs which are used as sources for system workflows.
      * @type {Record<string, Set<string>>}
      */
     packs: {
-      ancestry: new Set([SYSTEM.COMPENDIUM_PACKS.ancestry]),
-      background: new Set([SYSTEM.COMPENDIUM_PACKS.background]),
-      spell: new Set([SYSTEM.COMPENDIUM_PACKS.spell]),
-      talent: new Set([SYSTEM.COMPENDIUM_PACKS.talent]),
+      ancestry: game.settings.get("crucible", "ancestrySources"),
```
aaclayton, line 135: I think we need some fallback logic that if this setting `Set` is empty, it falls back to the default value. For example if `crucible.ancestrySources` is an empty set, that needs to mean "use the default ancestries" rather than "there are no ancestries".

aaclayton (changes requested):
Thanks for the contribution @danielrab - I have a few comments that I'd like to see addressed, and please also be sure to create a github issue that documents this intended change as per the Contribution Policy process (<https://github.com/foundryvtt/crucible/blob/master/CONTRIBUTING.md>).

crucible.mjs
aaclayton: Yeah, this would be my suggestion. Make the setting schema inherently tolerant, but:

1. Only present currently available packs in the UI selection
2. When setting the `CONFIG` values, make sure to only set the subset packs that currently exist.

crucible.mjs
danielrab: Just to be clear, we still want it to be in `crucible.CONFIG`, but we want to update it in `"setup"`, right? In that case I probably should initialize the sets as empty and then add to them based on the setting, that way if another module wants to modify the config in `"init"`, it won't be overriden.
Actually, about modules modifying things, I am not sure how a module is supposed to remove default options if we do this.

aaclayton:
Let's try and revisit this work so that it can be included in the `0.8.0` update. Will you have time to look at this in the next few weeks, or would you like me to take over from here?

danielrab:
sorry, I think I will not be working on this. If you want to do it yourself or anyone else wants to pick this up, feel free to.

PR #262, opened 2025-07-30
[#199] Archetypes can now provide initial equipment

### LINKED ISSUE #199: Allow Archetypes to provide initial equipment

roth-michael:
Closes #199 
A couple of adjacent questions/issues came up which I've added to my original comment on the issue.
Currently determining what's "valid equipment" based on a hardcoded check on item type, which is fine, but might make sense to have a field on items, depending on how often this determination comes up.

### DIFF
--- a/module/applications/sheets/item-archetype-sheet.mjs
+++ b/module/applications/sheets/item-archetype-sheet.mjs
@@ -76,6 +76,7 @@ export default class CrucibleArchetypeItemSheet extends CrucibleBackgroundItemSh
           img: item.img,
           description: item.system.description.public,
           tags: item.getTags(),
+          cssClass: [item.type, equipped ? "equipped" : ""].filterJoin(" "),
           quantity,
           equipped
         }
@@ -124,7 +125,7 @@ export default class CrucibleArchetypeItemSheet extends CrucibleBackgroundItemSh
 
   /**
    * Handle drop events for an equipment item added to this sheet.
-   * @param {DragEvent} event 
+   * @param {DragEvent} event
    * @returns {Promise<*>}
    */
   async #onDropEquipment(event) {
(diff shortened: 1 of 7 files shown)


module/applications/sheets/item-archetype-sheet.mjs
```diff
@@ -9,9 +9,35 @@ export default class CrucibleArchetypeItemSheet extends CrucibleBackgroundItemSh
   static DEFAULT_OPTIONS = {
     item: {
       type: "archetype"
+    },
+    actions: {
+      removeEquipment: CrucibleArchetypeItemSheet.#onRemoveEquipment
```
aaclayton, line 14: Let's add this handler to the upstream class `CrucibleActorDetailsItemSheet`. I anticipate we'll want Backgrounds to also provide equipment someday.

I think we'll only expose the data model and UX for equipment to the Archetype for now, but we can put the underlying infrastructure in place for any `CrucibleActorDetailsItemSheet` subclass to utilize.

module/models/item-archetype.mjs
```diff
@@ -22,6 +22,7 @@ export default class CrucibleArchetypeItem extends foundry.abstract.TypeDataMode
       }, {}), {validate: CrucibleArchetypeItem.#validateAbilities}),
       talents: new fields.SetField(new fields.DocumentUUIDField({type: "Item"})),
       skills: new fields.SetField(new fields.StringField({required: true, choices: SYSTEM.SKILLS})),
+      equipment: new fields.SetField(new fields.DocumentUUIDField({type: "Item"})),
```
aaclayton, line 25: I think we need a bit more than this because we need to support:
- Quantity (consider an assassin that should have 2x daggers)
- Equipped state (perhaps need to record whether the items should be initially equipped or not)

I think maybe something like:
```suggestion
      equipment: new fields.ArrayField(new fields.SchemaField({
        item: new fields.DocumentUUIDField({type: "Item"}),
        quantity: new fields.NumberField({required: true, nullable: false, integer: true, initial: 1}),
        equipped: new fields.BooleanField()
      })),
```
This would give us a bit of room to grow, if we need to configure any other properties about granted equipment in the future.

module/applications/sheets/item-archetype-sheet.mjs
```diff
@@ -28,16 +54,46 @@ export default class CrucibleArchetypeItemSheet extends CrucibleBackgroundItemSh
...
+      equipment: await this._prepareEquipment(),
+      equipmentPartial: this.constructor.INCLUDED_EQUIPMENT_TEMPLATE
     });
   }
 
   /* -------------------------------------------- */
 
+  /**
+   * Retrieve equipment and prepare for rendering.
+   * @returns {Promise<object[]>}
```
aaclayton, line 67: Would be good to get a bit more IDE typing support here:
```suggestion
   * @returns {Promise<{uuid: string, name: string, img: string, description: string, tags: object[]}[]>}
```

aaclayton (changes requested):
Nice one @roth-michael - I appreciate how you closely aligned this with the existing handling of talents. 

A couple comments for further iteration!

aaclayton (approved):

PR #266, opened 2025-08-05
Remove special Impetus logic in favor of `hastened` condition being granted

roth-michael:
I also added `startRound: this.round`, otherwise the existing effect (if there is one) would get removed anyway because of a 0-round duration. 
The actual solution to that might be different - impetus has a duration of 1 turn. Does that mean it should _expire_ at the end of 1 turn? The docs on `CrucibleActor##updateEndTurnEffects` say "Effects with a duration specified in Turns expire at the end of the Actor's turn once the duration has elapsed." Not sure if the duration should be counted as elapsed at the end of the combatant's turn. I suppose the question is actually two-fold:
1. Does "Turns" duration mean "After this combatant has had X turns" or "After X turns have passed"? My initial thinking was the latter, but I think currently it's implemented as the former (i.e. the only difference between Rounds and Turns is that Rounds expires at the _start_ of a turn, and Turns expires at the _end_ of a turn)
2. Does the turn that a "Turns"-duration effect is applied count as a turn? i.e. should a 1-Turn duration effect expire immediately after the turn ends?

### DIFF
--- a/module/documents/combat.mjs
+++ b/module/documents/combat.mjs
@@ -152,11 +152,12 @@ export default class CrucibleCombat extends foundry.documents.Combat {
         _id: impetusId,
         name: "Impetus",
         icon: "icons/magic/movement/trail-streak-zigzag-yellow.webp",
-        changes: [{key: "system.status.impetus", value: 1, mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE}],
+        statuses: ["hastened"],
         duration: {
           combat: this.id,
           rounds: 0,
-          turns: 1
+          turns: 1,
+          startRound: this.round
         }
       }
       if ( firstActor.effects.has(impetusId) ) await firstActor.updateEmbeddedDocuments("ActiveEffect", [impetus]);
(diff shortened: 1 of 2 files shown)


module/documents/combat.mjs
```diff
@@ -152,11 +152,12 @@ export default class CrucibleCombat extends foundry.documents.Combat {
...
         name: "Impetus",
         icon: "icons/magic/movement/trail-streak-zigzag-yellow.webp",
-        changes: [{key: "system.status.impetus", value: 1, mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE}],
+        statuses: ["hastened"],
         duration: {
           combat: this.id,
           rounds: 0,
-          turns: 1
+          turns: 1,
+          startRound: this.round
```
aaclayton, line 160: Impetus should have a duration specified in `rounds` which expire at the beginning of the Actor's turn on the subsequent round.

module/documents/combat.mjs
```diff
@@ -152,11 +152,12 @@ export default class CrucibleCombat extends foundry.documents.Combat {
         _id: impetusId,
         name: "Impetus",
         icon: "icons/magic/movement/trail-streak-zigzag-yellow.webp",
-        changes: [{key: "system.status.impetus", value: 1, mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE}],
+        statuses: ["hastened"],
         duration: {
           combat: this.id,
           rounds: 0,
```
aaclayton, line 158:
```suggestion
          rounds: 1,
```

module/documents/combat.mjs
aaclayton, line 160:
```suggestion
```

aaclayton (changes requested):

PR #278, opened 2025-08-15
[#142] Proposal: Add tooltip cards for Skills

### LINKED ISSUE #142: Add tooltip cards for Skills which reminds the reader what each skill does, which knowledges it provides, and other information

Plushtoast:
Fixes #142

Hi,
proposal for the skill card tooltip.

Currently looks like this:
grafik

Open questions:

- [ ] unclear where to put refactored prepareSkill method, should maybe be a static function in the actor data model?
- [ ] unclear if crucible.CONFIG.knowledge is knowledge which every character has and which should always be shown
- [ ] unclear if instead of filtering knowledge with actor.hasKnowledge just should take knowledge from the background (probably missing a collect knowledge function somewher
- [ ] could extend the method so it works without an actor for compendium/enricher

I stop here because I dont know if you like pull requests for topics you already assigned to yourself.
If do you consider to be missing?


aaclayton:
Part of the reason I didn't tag this topic for PR is that I have in my mind a planned refactor for skills that will probably interfere with some (much?) of this code. 

I think I'd like to do that refactor first and then figure out the right way to attach tooltips to the resulting behavior, so I'm going to mark this as blocked for a little while and we can revisit once that first step is done. Hope that makes sense!

aaclayton:
Will circle back to this later once Skills get their rework.

PR #279, opened 2025-08-15
Tidbits: Comply with action function signature

Plushtoast:
action handlers already include the target on the callback call

changes reflect function signature

not a big problem, but once I saw I wasnt able to unsee it anymore


aaclayton:
Thanks!

PR #283, opened 2025-08-20
[#282] Proposal add tags with tooltips to actor sheet effects for improved clarity

### LINKED ISSUE #282: Proposal: Add statuses of effects including tooltips to tags on actor sheet (effects tab)
Proposal:

Add statuses as tags to effects so their effect can be recalled by hovering the element within a tooltip.

Image

Image

Plushtoast:
Fixes #282 

grafik


aaclayton (approved):
Very nice! Easy win.

PR #298, opened 2025-08-23
[#289] Add data model & config for languages known

roth-michael:
Handles some of #289 
Also threw in the context prep in `CrucibleBaseActorSheet`. As-is, once a spot is determined for this to live, the multiselect could just be 
```hbs
{{formGroup fields.languages value=source.system.languages options=languages.options groups=languages.groups}}
```
Actually, should `languages` and `languageCategories` be adde to `crucible.CONFIG`?

### DIFF
--- a/module/applications/sheets/base-actor-sheet.mjs
+++ b/module/applications/sheets/base-actor-sheet.mjs
@@ -648,12 +648,9 @@ export default class CrucibleBaseActorSheet extends api.HandlebarsApplicationMix
     const categories = SYSTEM.ACTOR.LANGUAGE_CATEGORIES;
     const options = [];
     for ( const [value, {label, category}] of Object.entries(SYSTEM.ACTOR.LANGUAGES) ) {
-      const currOption = { value, label };
-      if ( category && categories[category] ) currOption.group = categories[category];
-      options.push(currOption);
+      options.push({value, label, group: categories[category]?.label});
     }
-    const groups = options.reduce((acc, {group}) => (group && !acc.includes(group)) ? acc.concat(group) : acc, []);
-    return { options, groups };
+    return options;
   }
 
   /* -------------------------------------------- */
--- a/module/config/actor.mjs
+++ b/module/config/actor.mjs
@@ -139,9 +139,16 @@ export const TRAVEL_PACES = freezeEnum({
 
 /**
  * Categories a language can (optionally) belong to
- * @type {Record<string, string>}
+ * @type {Record<string, {label: string}>}
  */
-export const LANGUAGE_CATEGORIES = {};
+export const LANGUAGE_CATEGORIES = {
+  nonSpoken: {
+    label: "LANGUAGE_CATEGORIES.NONSPOKEN"
+  },
+  spoken: {
+    label: "LANGUAGE_CATEGORIES.SPOKEN"
+  }
+};
 
 /**
  * Languages a creature can know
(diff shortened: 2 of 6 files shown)


module/config/actor.mjs
```diff
@@ -137,6 +137,22 @@ export const TRAVEL_PACES = freezeEnum({
   }
 });
 
+/**
+ * Categories a language can (optionally) belong to
+ * @type {Record<string, string>}
+ */
+export const LANGUAGE_CATEGORIES = {};
```
aaclayton, line 144: Let's put in an initial category, maybe something like "Spoken Languages" and we can put `common` inside it.

module/models/actor-base.mjs
```diff
@@ -111,6 +111,9 @@ export default class CrucibleBaseActor extends foundry.abstract.TypeDataModel {
     // Status
     schema.status = new fields.ObjectField({nullable: true, initial: null});
     schema.favorites = new fields.SetField(new fields.StringField({blank: false}));
+
+    // Languages
+    schema.languages = new fields.SetField(new fields.StringField({blank: false}));
```
aaclayton, line 116: I think I'd like to put this under `system.details.languages` instead of just `system.languages`

module/applications/sheets/base-actor-sheet.mjs
```diff
@@ -643,6 +644,20 @@ export default class CrucibleBaseActorSheet extends api.HandlebarsApplicationMix
 
   /* -------------------------------------------- */
 
+  #prepareLanguages() {
+    const categories = SYSTEM.ACTOR.LANGUAGE_CATEGORIES;
+    const options = [];
+    for ( const [value, {label, category}] of Object.entries(SYSTEM.ACTOR.LANGUAGES) ) {
+      const currOption = { value, label };
+      if ( category && categories[category] ) currOption.group = categories[category];
+      options.push(currOption);
```
aaclayton, line 653: Can be a bit more terse:
```suggestion
      options.push({value, label, group: categories[category]});
```

module/applications/sheets/base-actor-sheet.mjs
```diff
@@ -643,6 +644,20 @@ export default class CrucibleBaseActorSheet extends api.HandlebarsApplicationMix
...
+    }
+    const groups = options.reduce((acc, {group}) => (group && !acc.includes(group)) ? acc.concat(group) : acc, []);
+    return { options, groups };
```
aaclayton, line 656: We don't need to pass the `groups` parameter unless we need manual control over the order in which groups are listed
```suggestion
    return options;
```

module/config/actor.mjs
```diff
@@ -137,6 +137,22 @@ export const TRAVEL_PACES = freezeEnum({
   }
 });
 
+/**
+ * Categories a language can (optionally) belong to
+ * @type {Record<string, string>}
```
aaclayton, line 142: I would prefer we go with:
```suggestion
 * @type {Record<string, {label: string}>}
```
This gives us more room to grow in case we ever need other category metadata

module/config/actor.mjs
```diff
@@ -137,6 +137,22 @@ export const TRAVEL_PACES = freezeEnum({
...
+ */
+export const LANGUAGE_CATEGORIES = {};
+
+/**
+ * Languages a creature can know
+ * @type {Record<string, {label: string, category?: string}>}}
+ */
+export const LANGUAGES = {
+  common: {
```
aaclayton, line 151: A nice idea which could be pretty setting agnostic would be to also add `sign` as a language under a category for non-spoken languages.

aaclayton (changes requested):
Looks pretty good, a couple of things I'd suggest to iterate on

module/models/actor-base.mjs
aaclayton: Yeah, both

aaclayton (approved):

PR #308, opened 2025-08-24
[#291] Add 5e-compatible `[[/award]]` enricher for currency/milestones

### LINKED ISSUE #291: Add an `[[/award]]` enricher that is cross-compatible with the D&D5E system syntax that can be used to grant currency or milestone points from within journal  text content
- Dependant on #288

roth-michael:
Closes #291 
May be worth pulling out the "select a hazard target" logic into its own helper, if we decide to essentially re-use it for awards.
I can't quite get the styling right here, specifically for the currency icons. Part of that may be the icons themselves having _slightly_ variable amounts of empty space on each side, but
1. `margin-bottom: -1.5px` doesn't feel ideal and
2. Adjusting font size has mixed results (looks fine on lower sizes, the coins get comically small relative to the text on higher ones; but can't just do `background-size: cover` because of the afore-mentioned empty space)

I also nixxed 5e's `each` and `party`, favoring party treatment by default and figuring non-`each` treatment doesn't really make sense with milestone awards. Might be worth a checkbox for "divide currency each way" though.


lang/en.json
```diff
@@ -498,6 +498,15 @@
 "ADVANCEMENT.MilestoneAward": "Award Milestones",
 "ADVANCEMENT.ExpectedLevel": "Expected Level",
 
+"AWARD": {
+  "InvalidWarning": "Unable to parse {terms} as valid award terms.",
+  "Milestone": {
+    "one": "Milestone",
+    "other": "Milestones"
+  },
+  "NotGMWarning": "You must be a Gamemaster to use Award enrichers."
```
aaclayton, line 507: Let's anticipate likely additional warnings as we continue to develop this feature and go with:
```suggestion
  "WARNINGS": {
    "RequiresGM": "You must be a Gamemaster to use Award enrichers."
  }
```

module/enrichers.mjs
```diff
@@ -25,6 +25,12 @@ export function registerEnrichers() {
       pattern: /\[\[\/language (\w+)]]/g,
       enricher: enrichLanguage
     },
+    {
+      id: "dnd5eAward",
```
aaclayton, line 29: I would prefer just calling this `award` since I don't anticipate having a *different* enricher for crucible.

module/enrichers.mjs
```diff
@@ -102,6 +108,150 @@ function enrichDND5ESkill([match, terms]) {
...
+  const invalid = [];
+  let milestones;
+  for ( const part of terms.split(" ") ) {
+    if ( !part ) continue;
+    let [, amount, label] = part.match(pattern) ?? [];
+    label = label?.toLowerCase();
+    try {
+      new Roll(amount);
+      if ( label in crucible.CONFIG.currency ) currency[label] = amount;
+      else if ( ["xp", "milestone", "milestones"].includes(label) ) milestones = Number(amount);
```
aaclayton, line 132: I don't think we should support `xp` and we should ignore that term.

module/enrichers.mjs
```diff
@@ -102,6 +108,150 @@ function enrichDND5ESkill([match, terms]) {
...
+}
+
+/* -------------------------------------------- */
+
+async function onClickAward(event) {
+  event.preventDefault();
+  if ( !game.user.isGM ) return ui.notifications.warn("AWARD.NotGMWarning", { localize: true })
+
+  const terms = event.currentTarget.dataset.terms;
+  const { currency, milestones } = parseAwardTerms(terms);
```
aaclayton, line 197: We have already parsed award terms at the `enrichAward` phase, we shouldn't need to re-parse. We should have all the information we need as `dataset` attributes of the enriched tag.

We need to do some final verification here that the currency or milestones in the tag are still valid, but it shouldn't involve re-parsing raw string.

module/enrichers.mjs
```diff
@@ -102,6 +108,150 @@ function enrichDND5ESkill([match, terms]) {
...
+    entries.push(`
+      <span class="award-entry">
+        ${milestones} ${game.i18n.localize("AWARD.Milestone." + plurals.select(milestones))}
+      </span>
+  `);
+  }
+  // Return the enriched content tag
+  const tag = document.createElement("enriched-content");
+  tag.classList.add("award");
+  tag.dataset.terms = terms;
```
aaclayton, line 175: We should add more direct data here so we don't have to re-parse the string on click.

module/enrichers.mjs
```diff
@@ -102,6 +108,150 @@ function enrichDND5ESkill([match, terms]) {
...
+  const pattern = new RegExp(/^(.+?)(\D+)$/);
+  const currency = {};
+  const invalid = [];
+  let milestones;
+  for ( const part of terms.split(" ") ) {
+    if ( !part ) continue;
+    let [, amount, label] = part.match(pattern) ?? [];
+    label = label?.toLowerCase();
+    try {
+      new Roll(amount);
```
aaclayton, line 130: What is accomplished by constructing an un-evaluated `Roll` instance here?

module/enrichers.mjs
```diff
@@ -102,6 +108,150 @@ function enrichDND5ESkill([match, terms]) {
...
+  return { currency, milestones }
+}
+
+/**
+ * Enrich an Award with the format [[/award {...awards}]]
+ * @param {string} match
+ * @param {string} terms
+ */
+function enrichAward([match, terms]) {
+  const { currency, milestones } = parseAwardTerms(terms);
```
aaclayton, line 152: We need some error handling here if `parseAwardTerms` throws an Error

module/enrichers.mjs
```diff
@@ -102,6 +108,150 @@ function enrichDND5ESkill([match, terms]) {
...
+
+  for ( const [key, formula] of Object.entries(currency) ) {
+    const roll = await new Roll(`${formula}`).evaluate();
+    currency[key] = roll.total;
+  }
+
+  const netCurrency = crucible.api.documents.CrucibleActor.convertCurrency(currency);
+
+  // TODO: Consider pulling this logic out and using it for both hazard & award
```
aaclayton, line 206: It's a good thought, but fine to leave as a TODO for now in this PR.

module/enrichers.mjs
```diff
@@ -25,6 +25,12 @@ export function registerEnrichers() {
       pattern: /\[\[\/language (\w+)]]/g,
       enricher: enrichLanguage
     },
+    {
+      id: "dnd5eAward",
+      pattern: /\[\[\/award ([\w\s]+)]]/g,
```
aaclayton, line 30: I think the default behavior needs to be that the award is split (divided) among the party members and that we need to support an `each` keyword that is used to designate cases where every recipient receives the award.

For milestones awards, of course, `each` behavior is default.

styles/journal.less
```diff
@@ -236,6 +237,26 @@ enriched-content.hazard-check {
...
+      font-weight: 900;
+    }
+
+    .award-entry .currency {
+      display: inline-block;
+      height: var(--font-size-14);
+      width: var(--font-size-14);
+      background-position: center;
+      margin-bottom: -1.5px;
+    }
```
aaclayton, line 258: I think maybe we should try making the award `display: inline-flex` which would probably making this positioning easier without some ugly rules.

aaclayton (changes requested):
Looking good. Lots of comments for you on this one though!

module/enrichers.mjs
roth-michael: Admittedly took a fair bit of inspiration from 5e's handling of it - the idea is that it'll throw if someone puts some invalid term. Could do `if ( !Roll.validate(amount) ) throw new Error()` or some such, but that's minimally more overhead.

styles/journal.less
roth-michael: Having `.award-entry`'s `display: inline-flex` does remove the need for the `display: inline-block` display on the currency element, but still seem to need `width`,` height` (maybe), and `background-position`.

If we crop the included currency icons to be 13x13 rather than 24x24 (aka remove all the transparent bits on the right/left/top/bottom) then can get by with just `width` and `background-size: cover` (and it works in larger font sizes as well).

module/enrichers.mjs
aaclayton: hmm, yeah, let's go with `Roll.validate()` because it's more readable and explicit. We should probably only allow roll formulae for currency though. Awarding a random number of milestones doesn't make sense.

module/enrichers.mjs
```diff
@@ -100,6 +106,168 @@ function enrichDND5ESkill([match, terms]) {
...
+  } catch(err) {
+    currency = {};
+    milestones = 0;
+    each = false;
+    ui.notifications.error(err.message)
+  }
+  let entries = [];
+  for ( let [currencyKey, amount] of Object.entries(currency) ) {
+    const {label, icon, abbreviation} = crucible.CONFIG.currency[currencyKey];
+    const textOrIcon = icon ? `<i class="currency ${currencyKey}" data-tooltip aria-label="${label}" style="background-image: url(${icon});"></i>` : abbreviation;
```
aaclayton, line 165: It would be good to add classes for these like `<i class="currency gp">` etc... instead of using inline style rules.

module/enrichers.mjs
```diff
@@ -100,6 +106,168 @@ function enrichDND5ESkill([match, terms]) {
...
+  if ( !game.user.isGM ) return ui.notifications.warn("AWARD.WARNINGS.RequiresGM", { localize: true })
+
+  const { currency, milestones, each } = foundry.utils.expandObject({...event.currentTarget.dataset});
+
+  for ( const [key, formula] of Object.entries(currency) ) {
+    const roll = await new Roll(`${formula}`).evaluate();
+    currency[key] = roll.total;
+  }
+
+  let currencyEach = crucible.api.documents.CrucibleActor.convertCurrency(currency);
```
aaclayton, line 219: It would be good to do this conversion at the `enrichAward` stage rather than in the click handler, that way we only need one dataset attribute for currency (total converted currency) rather than multiple. Should simplify things.

aaclayton (changes requested):
Looking good! Just a few small lingering pieces of feedback. Thanks for your patience and iteration with me on this PR!

module/enrichers.mjs
roth-michael: While we do roll for every one, we still only actually allow numbers for milestones. But I can move the roll validate inside the `if ( label in crucible.CONFIG.currency )` block.

module/enrichers.mjs
roth-michael: When would be the ideal time to create these styles, do you think? We never freeze the currency config, but it's _probably_ a safe bet to do so any time post-init (5e does it in `setup` iirc).

module/enrichers.mjs
roth-michael: Had the same thought, but the enricher itself can have non-deterministic values for currencies. Evaluating at enrich-time _works_, but means that if you click the same enricher multiple times it'll have the same result, unless you re-render the journal page. If that's fine, can do it that way.

aaclayton (approved):
Good explanations. Approving and then will tweak further on my end!

PR #311, opened 2025-08-25
[#64] Add check to ensure no skipping training ranks

### LINKED ISSUE #64: Training talents do not correctly require that you satisfy the prerequisite prior level of training

roth-michael:
Closes #64 
e.g. you can't grab "Deception Journeyman" unless you've already got "Deception Novice"


aaclayton (approved):
Nice, thank you! Although I think this might end up going away in a future refactor.

PR #315, opened 2025-09-02
Allow "negative" currency `award`s, separate `milestone` enricher, add award/milestone summary chat cards

roth-michael:
Haven't tried to account for a mix of positive & negative values in an award, though it should "just work," if some maniac tries it. Depending on how much future improvements we put into negative award enrichers, it may make sense to make them their own "cost" enricher. 

Also fixed awarding milestones (unrelated to enricher).

### DIFF
--- a/module/models/actor-group.mjs
+++ b/module/models/actor-group.mjs
@@ -197,30 +197,48 @@ export default class CrucibleGroupActor extends foundry.abstract.TypeDataModel {
    * Award milestones to the members of this group.
    * @param {number} quantity
    * @param {string[]} recipientIds
+   * @param {boolean} [createMessage=true]
    * @returns {Promise<void>}
    */
-  async awardMilestones(quantity, recipientIds) {
+  async awardMilestones(quantity, recipientIds, createMessage=true) {
     recipientIds ||= Array.from(this.memberIds);
+    const recipientNames = [];
     const updates = recipientIds.reduce((arr, id) => {
       const actor = game.actors.get(id);
       if ( !actor ) return arr;
+      recipientNames.push(actor.name);
       const starting = actor.system._source.advancement.milestones;
-      const milestones = Math.max(starting + response.quantity, 0);
+      const milestones = Math.max(starting + quantity, 0);
       arr.push({_id: id, "system.advancement.milestones": milestones});
       return arr;
     }, []);
-    const groupMilestones = Math.max(this._source.advancement.milestones + response.quantity, 0);
+    const groupMilestones = Math.max(this._source.advancement.milestones + quantity, 0);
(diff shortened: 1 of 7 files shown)


module/models/actor-group.mjs
```diff
@@ -197,30 +197,48 @@ export default class CrucibleGroupActor extends foundry.abstract.TypeDataModel {
    * Award milestones to the members of this group.
    * @param {number} quantity
    * @param {string[]} recipientIds
+   * @param {boolean} [createMessage=true]
    * @returns {Promise<void>}
    */
-  async awardMilestones(quantity, recipientIds) {
+  async awardMilestones(quantity, recipientIds, createMessage=true) {
```
aaclayton, line 203: Let's put `recipientIds` and `createMessage` within an options object so there's room to grow the signature here:
```suggestion
  async awardMilestones(quantity, {createMessage=true, recipientIds}={}) {
```

module/models/actor-group.mjs
```diff
@@ -256,7 +274,7 @@ export default class CrucibleGroupActor extends foundry.abstract.TypeDataModel {
       content: `${quantityHTML.outerHTML}${recipientsHTML.outerHTML}`
     });
     if ( !response ) return;
-    await this.awardMilestones(response.quantity, response.recipients);
+    await this.awardMilestones(response.quantity, response.recipients, createMessage);
```
aaclayton, line 277:
```suggestion
    await this.awardMilestones(response.quantity, {...options, recipientIds: response.recipients});
```

module/models/actor-group.mjs
```diff
@@ -197,30 +197,48 @@ export default class CrucibleGroupActor extends foundry.abstract.TypeDataModel {
...
   /* -------------------------------------------- */
 
   /**
    * Provide the Gamemaster with a Dialog to award milestone points to the group.
+   * @param {number} [baseQuantity=1]       Quantity of milestones to pre-populate the dialog with
+   * @param {boolean} [createMessage=true]  Whether to create a chat message summarizing the award
    * @returns {Promise<void>}
    */
-  async awardMilestoneDialog() {
+  async awardMilestoneDialog(baseQuantity=1, createMessage=true) {
```
aaclayton, line 241: Pass any options on to the `awardMilestones` method.
```suggestion
  async awardMilestoneDialog(baseQuantity=1, options={}) {
```

styles/chat.less
```diff
@@ -53,6 +53,22 @@
   button.confirm {
     margin-top: 0.5rem;
   }
+
+  // Award currency summary
+  .award-chat {
```
aaclayton, line 58: It feels to me like we should be getting these styles for free. Is there a class or selector we can share with other contexts in which we present currency?

aaclayton (approved):

PR #316, opened 2025-09-02
[#313] Add guard to `CrucibleTalentTreeTalent##onPointerOver` to ensure tooltip only procs if talent tree is the topmost element under the pointer

### LINKED ISSUE #313: Talent tree tooltips appear when hovering on top of an application in front of the tree
This most obviously happens when the confirmation dialog overlaps one of the nodes that has a tooltip on it.

Example:

Image

roth-michael:
Closes #313 
`CrucibleTalentTreeNode##onPointerOver` has this check, which is why the same behavior isn't observed there.
Didn't add to `#onPointerOut` because hovering a window above the tree _should_ be grounds for de-tooltipping.

### DIFF
--- a/module/canvas/tree/talent-tree-talent.mjs
+++ b/module/canvas/tree/talent-tree-talent.mjs
@@ -85,6 +85,7 @@ export default class CrucibleTalentTreeTalent extends CrucibleTalentIcon {
   /* -------------------------------------------- */
 
   #onPointerOver(event) {
+    if ( document.elementFromPoint(event.globalX, event.globalY)?.id !== "crucible-talent-tree" ) return;
     event.stopPropagation();
     this.scale.set(1.2, 1.2);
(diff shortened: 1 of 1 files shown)


module/canvas/tree/talent-tree-talent.mjs
```diff
@@ -85,6 +85,7 @@ export default class CrucibleTalentTreeTalent extends CrucibleTalentIcon {
   /* -------------------------------------------- */
 
   #onPointerOver(event) {
+    if ( document.elementFromPoint(event.globalX, event.globalY)?.id !== "crucible-talent-tree" ) return;
```
aaclayton, line 88: Fine merging this for now, but I wonder if it's possible to retrieve this directly from `event.data.originalEvent` or something.

aaclayton (approved):

PR #321, opened 2025-09-05
Fixes #320

### LINKED ISSUE #320: Bug: 1-Handed weapons only have range of 1 feet and cannot attack from further than that
Currently 1-handed range weapons can not attack from a distance farther than 1feet:

Image

2-handed range weapons are fine though.

This bug happens on the master branch. not sure about the released version.

The reason seems to be that having a range weapon in one hand and nothing or a shield in the other hand always results in the strike being a meleeattack.

The commit 6931ccc5037ea7fbbbd39f4c8d0de968769b046d  "Update archetypes and equipment" has caused the fail.

Image

Plushtoast:
Fixes #320 

Changes:

* Moved weapons.dualWield after weapons.unarmed = because weapons.unarmed was undefined/wrong order
* Added check to see if one hand is unarmed/shield and the other is ranged so hasChoice is true (weapon.ranged && weapon.melee)


aaclayton (approved):
Thanks for the improvement!

PR #322, opened 2025-09-05
Add HitBox shader and layer classes

Feu-Secret:
https://github.com/user-attachments/assets/9d7d7bf8-847c-4645-aac5-5d3659301067


crucible.mjs
```diff
@@ -269,6 +271,21 @@ Hooks.once("init", async function() {
   CONFIG.debug.talentTree = false;
   CONFIG.debug.flanking = false;
   if ( crucible.developmentMode ) registerDevelopmentHooks();
+
+  // Add Hit Box layer
+  CONFIG.Canvas.layers.crucibleHitBoxes = {
```
aaclayton, line 276: What's your rationale for this being a new canvas layer rather than part of the existing grid layer?

module/canvas/hit-box.mjs
```diff
@@ -0,0 +1,76 @@
...
+  /* ---------------------------------------- */
+
+  /** @override */
+  _tearDown() {}
+
+  /* ---------------------------------------- */
+
+  /** @override */
+  render(renderer) {
+    if ( !this.worldVisible || this.worldAlpha <= 0 ) return;
```
aaclayton, line 42: Maybe some protection against being in a scene that is not `canvas.scene.useMicrogrid === true`?

module/canvas/hit-box.mjs
```diff
@@ -0,0 +1,76 @@
...
+  /* ---------------------------------------- */
+
+  /**
+   * Collect the tokens whose hitboxes should be drawn.
+   * @returns {Token[]|null} Visible tokens if at least one is controlled, otherwise null
+   */
+  _collect() {
+    const anyControlled = canvas.tokens.controlled.length > 0;
+    if ( !anyControlled ) return null;
+    return canvas.tokens.placeables.filter(t => t?.visible);
```
aaclayton, line 25: I don't really like that we loop twice here every frame, once to `filter` and again to iterate over the filtered list in `render`. 

If we want to design it this way it would be better for `_collect` to be a generator that yields each visible token one at a time or to just combine all of the logic together into a single loop in `render`.

module/canvas/shaders/hit-box-shader.mjs
```diff
@@ -0,0 +1,194 @@
...
+      vHalfSize = aHalfSize;
+      vColor = aColor * tint;
+    }
+    `;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static get batchFragmentShader() {
```
aaclayton, line 119: The initial animation is a good prototype, I'm sure we can make it nicer. I think the simplest thing to do as a starting point would be to add a little bit of color variation to it with some gradient of color rather than a solid color. Imagine the shape of the animation but with a sort of radial gradient that gives it a bit of subtle variation.

aaclayton (changes requested):
Awesome prototype Ludovic, thank you!

crucible.mjs
Feu-Secret: We don't override/monkeypatch anything. This is all using the Foundry API, putting the new layer in CONFIG, and having our class with 100% control on it. This is also very readable and easily maintained.

module/canvas/shaders/hit-box-shader.mjs
Feu-Secret: For the moment, I have not added an extra color. This is something we should discuss.

PR #353, opened 2025-10-26
[#209] Add damage type immunities

roth-michael:
Addresses at least part of #209 
Remaining implementation questions:
- Currently reduces damage to 1, should this change?
- How taxonomies should define immunity, what impact that should have on balance for vulnerabilities
- Finalize character sheet UI

And one nit:
- A Roll's `data.damage.resistance` being `Infinity` means when serialized it'll become `null`. Manageable by null coalescing it to `Infinity`, as done in `standard-check.mjs`, but worth keeping in mind in the future in case there's a more optimal solution (maybe adding a `immune` boolean to `data.damage` or something).

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -249,7 +249,7 @@
       },
       "immune": {
         "label": "Immunity",
-        "hint": "If immune to a damage type, damage of that type is reduced to 1."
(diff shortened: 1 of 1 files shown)


aaclayton:
Thanks for the contribution!

PR #354, opened 2025-10-26
[#330] Auto-resize tokens when importing non-Crucible scenes

### LINKED ISSUE #330: When importing a Scene from a Compendium pack that was not originally designed in the Crucible system, auto-migrate tokens to resize them for the Crucible microgrid
For example, a Scene created and packaged in `dnd5e` will contain a bunch of 1x1 tokens that should auto-convert to 4x4 tokens, etc...

roth-michael:
Closes #330 
Notes:
- Tokens aren't resized if the scene won't be auto-microgridded (i.e. if not square, if not ft, or if not 5 distance, no change).
- In order to maintain proper _locations_ of tokens, had to make other-system-1x1s into crucible 5x5s, not 4x4s, since in `CrucibleScene#prepareBaseData` the size of the grid is divided by 5, not 4.
- Had to do both `_preCreate` _and_ `_preUpdate`, the latter for direct json imports onto a scene, the former for compendium imports.
- Added another fallback of `this.document?.width` in `CrucibleTokenObject#getHitBoxData`, otherwise all actorless tokens would have their hitbox default to 4x4.

### DIFF
--- a/module/documents/scene.mjs
+++ b/module/documents/scene.mjs
@@ -99,7 +99,7 @@ export default class CrucibleScene extends Scene {
   /**
    * Determine whether input scene data will be made to use the microgrid automatically
    * @param {object} sceneData  The scene data to check
-   * @returns {Boolean}
+   * @returns {boolean}
    */
   static useMicrogrid(sceneData) {
     const g = sceneData.grid;


aaclayton (changes requested):
Thank you! A few questions

aaclayton (approved):
Thank you!

PR #358, opened 2025-10-27
[#345] Consolidate `defendXAttack` hooks to one `defendAttack` hook

### LINKED ISSUE #345: Consolidate the `defendWeaponAttack`, `defendSpellAttack`, and `defendSkillAttack` actor hooks under a shared `defendAttack` hook which is called for any sort of attack action.

roth-michael:
Closes #345 
As part of this, I moved the hooks of the affected talents which weren't currently in `hooks/talent.mjs` into there. Left the hooks on for Protective Mirage, though. Re-exported lv-6 Duurath because he had the old Testudo. 
Only one legacy hook remains in the pack data, and it's inside the playtest 1 adventure (it's the aforementioned Testudo on Duurath) so I figured not worth touching currently, especially since it won't cause any issues in functionality.

### DIFF
--- a/_source/pregens/Duurath_iLUfMcYjAbZFY4CJ.yml
+++ b/_source/pregens/Duurath_iLUfMcYjAbZFY4CJ.yml
@@ -1024,45 +1024,6 @@ items:
     ownership:
       default: 0
     _key: '!actors.items!iLUfMcYjAbZFY4CJ.bulwark000000000'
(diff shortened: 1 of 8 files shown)


aaclayton (approved):
Great stuff, thank you! I have 2 on the Ember side that also need to change. I'll update them now.

PR #361, opened 2025-10-28
[#209] Allow taxonomies to grant damage type immunity

roth-michael:
Continues iteration on #209 
As part of this approach, the data model for `CrucibleTaxonomyItem` has changed, `resistances` going from an object with keys of damage IDs and values of numbers to an object with keys of damage IDs and values of `{value: theOldValue, immune: boolean}`. I believe my change to `CrucibleTaxonomyItem#migrateData` (& changes elsewhere) should be sufficient to address that.

### DIFF
--- a/module/applications/sheets/item-taxonomy-sheet.mjs
+++ b/module/applications/sheets/item-taxonomy-sheet.mjs
@@ -40,7 +40,10 @@ export default class CrucibleTaxonomyItemSheet extends CrucibleActorDetailsItemS
       })),
       immunities: {
         field: new SetField(new StringField({choices: SYSTEM.DAMAGE_TYPES}), {label: "TAXONOMY.FIELDS.resistances.immune.label"}),
-        immune: Object.entries(context.source.system.resistances).filter(([_, {immune}]) => immune).map(([id]) => id)
+        immune: Object.entries(context.source.system.resistances).reduce((arr, [k, {immune}]) => {
+          if ( immune ) arr.push(k);
+          return arr;
+        }, [])
       }
     });
   }


templates/sheets/item/taxonomy-config.hbs
```diff
@@ -33,10 +33,16 @@
         <div class="form-fields">
             {{#each resistances as |r|}}
             <label>
-                {{formInput r.field value=r.value type="number"}}
+                {{formInput r.field value=r.value disabled=(eq r.value 4) type="number" dataset=(object damageType=r.id)}}
```
aaclayton: I don't love using these logical helpers in the HBS files and would prefer that we prepare the attributes we need for rendering as part of the `resistances` array that is passed into the template.
```suggestion
                {{formInput r.field value=r.value type="number" dataset=r.dataset disabled=r.disabled}}
```

module/applications/sheets/item-taxonomy-sheet.mjs
```diff
@@ -29,11 +29,15 @@ export default class CrucibleTaxonomyItemSheet extends CrucibleActorDetailsItemS
         value: context.source.system.abilities[ability.id]
       })),
       resistances: Object.values(SYSTEM.DAMAGE_TYPES).map(damage => ({
-        field: context.fields.resistances.fields[damage.id],
+        field: context.fields.resistances.fields[damage.id].fields.value,
         id: damage.id,
         label: damage.label,
-        value: context.source.system.resistances[damage.id]
-      }))
+        value: context.source.system.resistances[damage.id].immune ? 4 : context.source.system.resistances[damage.id].value
```
aaclayton: Prepare other things we need here: `disabled`, `dataset`

module/applications/sheets/item-taxonomy-sheet.mjs
```diff
@@ -74,6 +81,27 @@ export default class CrucibleTaxonomyItemSheet extends CrucibleActorDetailsItemS
...
+   * disabling the corresponding inputs and overriding their visual value as 4.
+   * @param {string[]} selected The ids of the damage types which are currently selected as immunities
+   */
+  #handleImmunities(selected) {
+    const resistances = this.element.querySelector(".resistances");
+    const inputs = resistances.querySelectorAll("input[type=number]");
+    for ( const input of inputs ) {
+      if ( selected.includes(input.dataset.damageType) ) {
+        input.disabled = true;
+        input.value = 4;
```
aaclayton: I'm thinking now that I'm seeing this, that if we are disabling the input regardless, maybe we can visually improve it by swapping the input type to `"text"` and displaying the value as `&infin;`? Instead of displaying it as `4` which is not as intuitive.

Issue #387, opened 2025-11-08
Implement Counterspell workflow to attempt negation of the previously cast Spell. Preferring an unconfirmed spell but potentially working for a confirmed one if no unconfirmed spell is available.


aaclayton:

roth-michael:
1. In the case that it's an unconfirmed spell, would confirming a successful Counterspell have any mechanical impact on that card/caster?
2. If it were a confirmed one, should confirming a successful Counterspell then reverse the cast spell?

I'm picturing "yes" to (2), and that the "confirm" button doesn't get rendered in the case that the spell in question has been confirmed successfully cancelled. This does make the assumption that countered spells do not result in the loss of their action/focus costs.

PR #393, opened 2025-11-10
[#387] Counter Spell workflow part 1

roth-michael:
Begins to address #387 
Currently, just adds a `CrucibleChatMessage.getLastAction`, which iterates backwards through `game.messages` until there's a valid action-linked message and returns the action in question if found, and adds a hook for `counterSpell` which prevents usage if the targeted token's actor is not responsible for the most recent action, or if that action isn't a spell.

### DIFF
--- a/module/hooks/action.mjs
+++ b/module/hooks/action.mjs
@@ -152,8 +152,8 @@ HOOKS.clarifyIntent = {
 HOOKS.counterSpell = {
   async preActivate(targets) {
     const target = targets[0];
-    const lastAction = ChatMessage.implementation.getLastAction();
-    if ( !lastAction || (target?.actor !== lastAction.actor) || !lastAction.tags.has("spell") ) {
+    const lastAction = ChatMessage.implementation.getLastAction({actor: target?.actor});
+    if ( !lastAction || !lastAction.tags.has("spell") ) {
       throw new Error("You can only counterspell the caster of the last action, and that action must be a spell!")
     }
   }


module/documents/chat-message.mjs
```diff
@@ -55,6 +55,20 @@ export default class CrucibleChatMessage extends ChatMessage {
     await CrucibleAction.confirmMessage(this, {action});
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Returns the most recent action (confirmed or otherwise) in the chat log
+   * @returns {CrucibleAction|null}
+   */
+  static getLastAction() {
```
aaclayton: Good start, I suggest adding some parameter options to this. some possibilities include:
* `{boolean} [confirmed]` - require the action to be confirmed (true) unconfirmed (false) or either (undefined)
* `{string|Actor} [actor]` - An Actor (or actor ID) who was the person to perform the action

Maybe others in the future? Signature like:
```suggestion
  static getLastAction({confirmed, actor}={}) {
```

module/documents/chat-message.mjs
```diff
@@ -55,6 +55,20 @@ export default class CrucibleChatMessage extends ChatMessage {
...
+    for ( const message of game.messages.contents.toReversed() ) {
```
aaclayton: `game.messages.contents.toReversed()` creates the array twice which is a bit wasteful. Better to create it once and then iterate backwards:

```suggestion
    const messages = game.messages.contents;
    for ( let i=messages.length-1; i>=0; i-- ) {
      const message = messages[i];
```

PR #422, opened 2025-11-14
[#419, #420] Successful counterspell confirmation flags & confirms the targeted message, with any damage/healing & effects ignored

### LINKED ISSUE #419: When a successful Counterspell is confirmed, modify the result of the target spell (Chat Message) to record that it was counterspelled and confirm it.

### LINKED ISSUE #420: When a Spell that was counterspelled is Confirmed, apply its effects without dealing any damage or causing any active effects.

roth-michael:
Closes #419 
Reversing a successful counterspell also flags (with false) and reverses the targeted message. The flag in question is `crucible.isCounterspelled` (to be made use of in the PR that covers #420).
A successful counterspell message can't be confirmed if targeted spell message is confirmed, and can't be reversed if the targeted spell message is unconfirmed. Either will result in a warning notification.
New stuff:
- If any `confirm` hook throws an error, action confirmation will end early. If that's being done via the chat card, the message will return to its previous confirmation state.
- Added `CrucibleAction#message` - this'll only be populated for actions which have been generated via `CrucibleAction.fromChatMessage`, and in that instance it'll provide a way to reference the originating message.

Also fixed a bug where non-dice-having gestures (aspect, creation, etc) weren't rolling the attack with counterspell.

Closes #420 
Any of `health`, `wounds`, `morale`, `madness` in `outcome.resources` are deleted, unless any of them are listed as a _cost_ for the action itself & the outcome has `self === true`, in which case that cost is used (this way, blood magic works just fine). `outcome.effects` is set to an empty array. Wasn't sure whether it made sense to block `outcome.actorUpdates` - I went with no.

Also added some `?.`s to account for the fact that iconic spells don't necessarily have a specific single gesture & rune to counter. Currently, no boons are afforded regardless of choice for such a spell.

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -1671,6 +1671,7 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
    * @param {object} options              Context options for ChatMessage creation
    * @param {boolean} options.confirmed   Were the outcomes auto-confirmed?
    * @returns {Promise<ChatMessageData>}
+   * @protected
    */
   async _prepareMessage(targets, {confirmed}={}) {
 
@@ -1736,10 +1737,8 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
    * @param {boolean} [options.confirmed]           Were the outcomes auto-confirmed?
    * @returns {Promise<ChatMessage>}              The created ChatMessage document
    */
-  async toMessage(targets, {confirmed=false, ...options}={}) {
-    const messageData = await this._prepareMessage(targets, {confirmed});
-
-    // Create chat message
+  async toMessage(targets, options={}) {
+    const messageData = await this._prepareMessage(targets, options);
     return ChatMessage.create(messageData, options);
   }
 


aaclayton:
This generally looks good. I'm going to hold it for the next release in the interest of time, but let's keep pushing on!

module/models/action.mjs
```diff
@@ -1417,9 +1423,12 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
       } catch(err) {
         console.error(new Error(`"${this.id}" action confirmation failed`, {cause: err}));
+        return false;
       }
     }
 
+    const isNegated = this.message?.getFlag("crucible", "isCounterspelled");
+
     // Additional Actor-specific consequences
     this.actor.onDealDamage(this, this.outcomes);
```
aaclayton: An action being negated should probably prevent it from triggering `onDealDamage` workflows.

module/models/action.mjs
```diff
@@ -1716,7 +1733,10 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
   static async confirmMessage(message, {action, reverse=false}={}) {
     action ||= this.fromChatMessage(message);
     await message.update({flags: {crucible: {confirmed: !reverse}}}); // Mark the message as confirmed *first*
-    await action.confirm({reverse}); // Then perform the confirmation effects
+    const confirmed = await action.confirm({reverse}); // Then perform the confirmation effects
+
+    // Un-flag message if confirmation/reversal didn't go through
+    if ( !confirmed ) await message.update({flags: {crucible: {confirmed: reverse}}});
```
aaclayton: A few thoughts here: 

1. I don't recall exactly what the requirement is for the message to be marked as confirmed *before* `CrucibleAction#confirm` is called. Is that a real requirement? Could we just wait until after we know whether confirmation was successful and then update the message?
2. I think probably `CrucibleAction#confirm` should take over responsibility for updating the `ChatMessage` flag. We have now given `CrucibleAction` awareness of its `message` property, and things like a programmatic call to `CrucibleAction#confirm` should also update the confirmation status of a corresponding chat message document if there is one. 

We could handle that as part of this PR, or as a follow up. I'll create a separate issue in case its better to handle this as a follow-up. https://github.com/foundryvtt/crucible/issues/449

module/models/action.mjs
```diff
@@ -1762,6 +1782,9 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
       outcome.rolls = outcomeRolls.map(i => message.rolls[i]);
       action.outcomes.set(outcome.target, outcome);
     }
+
+    // Add reference to originating message
+    action.message = message;
```
aaclayton: Let's pass the `message` in as part of `actionContext` to the constructor instead of assigning a reference post-construction and post-preparation.

aaclayton (changes requested):
Overall looking really good. I flagged some things I'd like to see updated or improved before we merged this in. I also identified a couple of additional issues that could be follow-ups.

aaclayton (changes requested):

PR #432, opened 2025-11-16
Removed the freehand action tag and switched usages to use cost.hands = 1

OsiJr:
Solves #428 (at least so it's clear what needs to be required to change for that conversation)

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -61,7 +61,7 @@ import CrucibleActionConfig from "../applications/config/action-config.mjs";
  * @property {number} action                The cost in action points
  * @property {number} focus                 The cost in focus points
  * @property {number} heroism               The cost in heroism points
- * @property {number} [hands]               A number of free hands required
+ * @property {number} [hands=0]             A number of free hands required
  * @property {boolean} [weapon]             Is the equipped weapon's action point cost added to this action's cost?
  */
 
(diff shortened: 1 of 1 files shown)


module/models/action.mjs
```diff
@@ -237,6 +239,7 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
         action: new fields.NumberField({required: true, nullable: false, integer: true, initial: 0}),
         focus: new fields.NumberField({required: true, nullable: false, integer: true, initial: 0}),
         heroism: new fields.NumberField({required: true, nullable: false, integer: true, initial: 0}),
+        hands: new fields.NumberField({required: false, nullable: false, integer: true, min: 0, initial: undefined}),
```
aaclayton: I think we should do `initial: 0` here such that most actions don't require a free hand unless otherwise specified. Should also probably set `max: 2` although there might be special exceptions?
```suggestion
        hands: new fields.NumberField({required: false, nullable: false, integer: true, initial: 0, min: 0, max: 2}),
```

module/models/action.mjs
```diff
@@ -60,7 +60,9 @@ import CrucibleActionConfig from "../applications/config/action-config.mjs";
  * @typedef {Object} ActionCost
  * @property {number} action                The cost in action points
  * @property {number} focus                 The cost in focus points
+ * @property {number} heroism               The cost in heroism points
  * @property {number} [hands]               A number of free hands required
```
aaclayton:
```suggestion
 * @property {number} [hands=0]             A number of free hands required
```

PR #442, opened 2025-12-02
[#418] Activation tags in header of action use dialog can gain unmet class

### LINKED ISSUE #418: Have activation tags gain the `unmet` class in the action use header if necessary
For instance, if casting a spell and selecting a 2-handed gesture with only one hand free, or selecting an inflection which would increase the action or focus cost beyond what the actor currently possesses. I think this would only come up for the spellcasting dialog, since everything else has a static cost known prior to use and therefore will be blocked outright in `canUse`.

roth-michael:
Closes #418 
I don't love everything about this solution, but not sure which direction would make the most sense to go instead. Initially I was thinking move from `Record<string, string>` to `Record<string, object>` with the object containing either a function to validate prerequisites or a style class, tooltip, etc. But thought that'd be overkill, so instead just added a list of unmet tag ids. Considered a handlebars helper for "does list X include entry Y" but for the time being just mapped that list to an object instead. Maybe that makes sense, and from the jump it should be defined as an object with tag ids as keys and tooltips explaining why they're invalid as values.

### DIFF
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -230,13 +230,20 @@ Hooks.once("init", async function() {
 
   // Register Handlebars helpers
   Handlebars.registerHelper({
-    "crucibleTags": (tags) => {
+    "crucibleTags": (tags, options={}) => {
+      const {enclosed=true, additionalClasses=null, tagType=null} = options.hash ?? {};
       const tagSpans = Object.entries(tags).map(([id, tag]) => {
-        const classes = `tag${tag.unmet ? " unmet" : ""}`;
+        let classes = "tag";
+        if ( tag.unmet ) classes += " unmet";
+        if ( tag.cssClasses ) classes += ` ${foundry.utils.escapeHTML(tag.cssClasses)}`;
         const label = foundry.utils.escapeHTML(tag.label ?? tag);
-        return `<span class="${classes}" data-tag="${id}">${label}</span>`;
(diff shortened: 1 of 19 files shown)


roth-michael:
Yeah, I can give that a go.

crucible.mjs
```diff
@@ -228,6 +228,18 @@ Hooks.once("init", async function() {
   // Rich Text Enrichers
   registerEnrichers();
 
+  // Register Handlebars helpers
+  Handlebars.registerHelper({
+    "crucibleTags": (tags) => {
```
aaclayton: I think if the helper operates on the entire `tags` array then it should (perhaps optionally, with a provided parameter) take responsibility for also writing the enclosing `<div class="tags">` element.

roth-michael: Actually, I think not. Most of the time it's just "tags" so having to write that out each time would be mildly annoying. Probably named args instead. A boolean for `enclosed` (default true) and an `additionalClasses` to specify additional classes on the enclosing div.

roth-michael:
Alright, think the helper's in a solid enough state. There are some places where we can't use it because there's additional `dataset` stuff set on each tag, which we could maybe address, but not sure if it's worth it. 

Also, currently the `--tag-color` variable is _only_ described here. Could (should?) add to the actual css for tags, maybe migrate the `unmet` styling to instead specify it as well.

PR #455, opened 2025-12-08
Comprehensive localization

### LINKED ISSUE #444: [Localization] Missing translation - Character creation
Hello!
Below I will post all the keys that I haven't been able to translate in Character Creation. Each group of missing translations will have its own post.

### LINKED ISSUE #445: [Localization] Missing translation - Character sheet
Hello!
Below I will post all the keys that I haven't been able to translate in Character Sheet. Each tab will have its own post.

### LINKED ISSUE #446: [Localization] Missing translation - Actions
Hello!
Below I will post all the keys that I haven't been able to translate in Actions sheet.

### LINKED ISSUE #447: [Localization] Missing translation - Backgrounds
Hello!
Below I will post all the keys that I haven't been able to translate in Backgrounds sheet.

### LINKED ISSUE #456: [Localization] Missing translation - System
Hello!
Below I will post all sytem keys that I haven't been able to translate.

roth-michael:
Closes #444, closes #445, closes #446, closes #447 
Mostly closes #456 
Made a best guess at what "made sense" for the paths of certain localizations based on what already existed. Happy to reorganize/reformat if desired.
Some remaining questions/points:
- Tooltips are provided in context for the hero creator's exit/restart/complete, but not used in the hbs
- I've used "TALENT.LABELS.Points.other" ("Points") for a place where "Points" refers to _ability_ points. Don't think this will ever be a meaningful difference, but not sure if I should do a separate localization just for it.
- There are some places where size and stride are shown as "Size 5ft" and others as just "Size 5" - should these be changed to stick to just one convention, and if so, which?
- Should abbreviations like "3A" (3 action points) be localized?
- Should thrown errors be localized? They may show up as tooltips/notification banners.
- There are a couple of references to "Protagonist" - am I right that that's an outdated name for "Hero"?
- For `SYSTEM.ACTION.TARGET_SCOPES` (an Enum) and `SYSTEM.DICE.checkDifficulties` (a `Record<number, string>`) there's no existing mechanism for prelocalization. Currently I'm leaving both in their unlocalized form and localizing during any relevant renders, but I'm wondering whether we should add a mechanism for this.
- Consumable config has `{{formInput X}} <label>of</label> {{formInput Y}}`, I'm not sure whether that'll be appropriate for all language patterns.
- There are a couple of `Effect: <span.../>` strings I'm not 100% sure how to handle. Maybe just a localization key with `"Effect: {whatever}"` and doing `{{{localize}}}` instead of `{{localize}}`? Only because `X: Y` may not be appropriate in all languages.


aaclayton:
Thanks very much for this! it's a big one to review, in future might be nice to have a sequence of smaller PRs, but I'll get through it sometime soon!

roth-michael:
Alright - cleaned this up into #458 #459 #460 and #461 - as they're merged in they'll probably need slight conflict resolution because they modify the same areas of the `en.json`, but that'll be easy.

PR #458, opened 2025-12-10
Creation-related localization

### LINKED ISSUE #444: [Localization] Missing translation - Character creation

roth-michael:
Closes #444 
Notes:
- Tooltips are provided in context for the hero creator's exit/restart/complete, but not used in the hbs
- I've used "TALENT.LABELS.Points.other" ("Points") for a place where "Points" refers to ability points. Don't think this will ever be a meaningful difference, but not sure if I should do a separate localization just for it.
- ~~There are some places where size and stride are shown as "Size 5ft" and others as just "Size 5" - should these be changed to stick to just one convention, and if so, which?~~

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -189,6 +189,12 @@
 },
 
 "ACTOR": {
+  "ACTIONS": {
+    "AppliedDetailItem": "Applied the {name} {type} to {actor}.",
+    "ClearedDetailItem": "Removed the {type} from {actor}.",
+    "DecreaseAbility": "Decrease Ability Score",
+    "IncreaseAbility": "Increase Ability Score"
+  },
   "CREATION": {
     "AbandonContent": "Discard creation progress and exit the creator?",
     "AbandonTitle": "Abandon Creation Process?",
@@ -380,16 +386,12 @@
   }
 },
 
-"ACTOR.AppliedDetailItem": "Applied the {name} {type} to {actor}.",
-"ACTOR.ClearedDetailItem": "Removed the {type} from {actor}.",
 "ACTOR.KnowledgeSpecific": "Knowledge: {knowledge}",
 "ACTOR.LanguageSpecific": "Language: {language}",
 "ACTOR.LevelSpecific": "Level {level}",
 "ACTOR.ResistanceSpecific": "Resistance: {resistance}",
 "ACTOR.SizeSpecific": "Size {size}",
 "ACTOR.StrideSpecific": "Stride {stride}",
-"ACTOR.TooltipDecreaseAbility": "Decrease Ability Score",
-"ACTOR.TooltipIncreaseAbility": "Increase Ability Score",
 "ACTOR.VulnerabilitySpecific": "Vulnerability: {vulnerability}",
 
 "ANCESTRY": {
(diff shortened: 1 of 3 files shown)


lang/en.json
```diff
@@ -366,8 +366,29 @@
   }
 },
 
+"ACTOR.AbandonCreationContent": "Discard creation progress and exit the creator?",
```
aaclayton: I think it would be good to put creation-specific strings under an `ACTOR.CREATION` namespace rather than having a ton of keys in the base `ACTOR` namespace.

lang/en.json
```diff
@@ -1364,6 +1385,10 @@
   },
   "PointsAvailable": "Points Available",
   "Purchase": "<p>Spend 1 Talent Point to purchase <strong>{name}</strong>?</p>",
+  "PurchaseNode": "Spend talent point to purchase empty node \"{node}\"?",
+  "PurchaseNodeReverse": "Remove point spent on empty node \"{node}\"?",
+  "PurchaseNodeTitle": "Purchase Talent Node?",
```
aaclayton: Let's put strings that are about UX actions under an `"ACTIONS"` namespace.

roth-michael: Looking at core's i18n, feels evident you meant the former. What determines, for instance, whether it should be `TALENT.ACTIONS.PurchaseNode` vs `TALENT.ACTIONS.PURCHASE_NODE`? In core for instance there's both `CARDS.ACTIONS.Create` and `AMBIENT_SOUND.ACTIONS.CREATE`.

templates/sheets/creation/background.hbs
```diff
@@ -44,23 +44,23 @@
...
+        <p id="ability-points"><span class="points">{{abilities.points.pool}}</span> {{localize "TALENT.LABELS.Points.other"}}</p>
         {{#each abilities.abilities as |ability|}}
         <div class="ability flexrow" data-ability="{{ability.id}}">
             <h4>{{ability.label}}</h4>
             <span class="score hex {{ability.group}}">{{ability.total}}</span>
             <span class="increases">{{ability.increases}}</span>
             <div class="buttons flexcol">
                 <button type="button" class="plain icon fa-solid fa-caret-up" data-action="abilityIncrease"
-                        data-tooltip="Increase Ability Score" {{#unless ability.canIncrease}}disabled{{/unless}}>
+                        data-tooltip="{{localize "ACTOR.TooltipIncreaseAbility"}}" {{#unless ability.canIncrease}}disabled{{/unless}}>
```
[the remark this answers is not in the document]
roth-michael: Should this be the case for all tooltips, or only those on buttons and the like?

aaclayton (changes requested):
Two more small changes

templates/sheets/creation/background.hbs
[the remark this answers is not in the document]
aaclayton: Bigger priority for interactive elements like buttons.

aaclayton (approved):

PR #460, opened 2025-12-10
Item related localization

### LINKED ISSUE #447: [Localization] Missing translation - Backgrounds

roth-michael:
Closes #447 and pieces of #445 
Notes:
- Consumable config has `{{formInput X}} <label>of</label> {{formInput Y}}`, I'm not sure whether that'll be appropriate for all language patterns.
- There are a couple of `Effect: <span.../>` strings I'm not 100% sure how to handle. Maybe just a localization key with `"Effect: {whatever}"` and doing `{{{localize}}}` instead of `{{localize}}`? Only because `X: Y` may not be appropriate in all languages.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -941,8 +941,19 @@
     "SCALED_PRICE": "Scaled Price"
   },
   "PROPERTIES": {
-    "INVESTMENT": "Investment",
-    "STACKABLE": "Stackable"
+    "Armor": "{armor} Armor",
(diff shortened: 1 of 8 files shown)


aaclayton (changes requested):
Small request and then I think this one is ready to go.

PR #480, opened 2025-12-15
[#467] Add `CrucibleActor#syncIconicSpells`

### LINKED ISSUE #467: Add a `CrucibleActor#syncIconicSpells` method that mirrors the behavior of `CrucibleActor#syncTalents`, replacing known iconic spells with updated versions from compendium packs.

roth-michael:
Closes #467 
Also includes this new function in the main crucible method `syncTalents`, and moves the update of `_stats.systemVersion` from `CrucibleActor#syncTalents` to `syncTalents`.

In theory don't really need the `toCreate` and `toDelete`, but if even we have iconic spell id migrations the way we do for talents currently it'll come in handy.


aaclayton (commented):
Looks good, but gonna hold off on this change until after the patch in case it introduces any unexpected migration issues

PR #494, opened 2025-12-20
Create `crucible` base AE type, move many flags into `system`

roth-michael:
Should mechanically be all working. ~~Still need to add localization, though the only spot I can think where any of it would show up is the to-be-added `TYPES.ActiveEffect.crucible` in the AE Config window title.~~ Some basic localization too.


aaclayton (approved):
Nice work, let's give it a spin

PR #532, opened 2025-12-28
[#527] Ensure window minimization is `await`ed before allowing template placement

### LINKED ISSUE #527: Placing a template too fast (while the action use dialog is still transitioning to minimized) can get the dialog stuck in a weird half-way state

roth-michael:
Closes #527. Root issue here may be an issue of calling `maximize` on the action use dialog while it's mid-`minimize`, but this solution feels appropriate regardless.

### DIFF
--- a/module/dice/action-use-dialog.mjs
+++ b/module/dice/action-use-dialog.mjs
@@ -247,7 +247,7 @@ export default class ActionUseDialog extends StandardCheckDialog {
         minimizedWindows.push(app);
       }
     }
-    await Promise.all(minimizedWindows.map(app => app.minimize()));
+    await Promise.allSettled(minimizedWindows.map(app => app.minimize()));
 
     // Store preview template data
     this.#targetTemplate = {


module/dice/action-use-dialog.mjs
```diff
@@ -239,16 +239,15 @@ export default class ActionUseDialog extends StandardCheckDialog {
...
         minimizedWindows.push(app);
       }
     }
     for ( const app of foundry.applications.instances.values() ) {
       if ( !app.minimized ) {
-        app.minimize();
         minimizedWindows.push(app);
       }
     }
+    await Promise.all(minimizedWindows.map(app => app.minimize()));
```
aaclayton: Maybe let's go with `allSettled` since it's not *critical* that all minimizations succeed, in case one fails for some reason.
```suggestion
    await Promise.allSettled(minimizedWindows.map(app => app.minimize()));
```

PR #533, opened 2025-12-29
Allow editing an Actor's Size Bonus via a dialog opened by clicking the hex

### LINKED ISSUE #469: Allow clicking on an Actor's Size on the character sheet in order to edit `system.movement.sizeBonus` via a simple dialog

Eldritch-Oliver:
This closes #469 and provides a base framework for handling #470 

Notes for reviewing:
- I made the hexagon as accessible as possible without being massively invasive to the styling (I would've preferred to make it a `<button>` but that caused a bunch of sizing issues, so `role="button" tabindex="0"` is good enough)
- Used `toFormGroup` on the schema field to keep it as concise as possible (even though it pained me from an accessibility standpoint to do so since that output is quite inaccessible)
- I also noticed that `sizeBonus` is sort of implied that it's minimum should be a negative value one less than the absolute `baseSize` (`(-1 * baseSize) + 1`), which is what the slightly odd `min` calculation is for the input configuration
- I would like to pull the handler method into the actor similar to how `_viewDetailItem` is essentially proxied from the sheet, is this a bad idea?

Example Dialog:
image

### DIFF
--- a/styles/elements.less
+++ b/styles/elements.less
@@ -15,6 +15,6 @@ crucible-currency {
   }
 }
 
-[data-action] {
+.crucible.sheet [data-action] {
   cursor: pointer;
 }


module/applications/sheets/adversary-sheet.mjs
```diff
@@ -91,6 +92,28 @@ export default class AdversarySheet extends CrucibleBaseActorSheet {
...
+   */
+  static async #onEditSize() {
+    const schema = this.actor.system.schema;
+    const { baseSize } = this.actor.system.movement;
+
+    const minSize = (baseSize * -1) + 1; // prevents sizes <= 0
+    const bonusField = schema.getField(`movement.sizeBonus`).toFormGroup({}, { min: minSize });
+
+    const formData = await foundry.applications.api.DialogV2.input({
+      window: {title: "Edit Movement: " + this.actor.name},
```
aaclayton: Localization

module/applications/sheets/hero-sheet.mjs
```diff
@@ -159,6 +160,28 @@ export default class HeroSheet extends CrucibleBaseActorSheet {
...
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Handle click actions to edit size bonuses
+   * @this {HeroSheet}
+   * @returns {Promise<void>}
+   */
+  static async #onEditSize() {
```
aaclayton: Instead of duplicating this handling on both sheets, I think we should put it on the `CrucibleBaseActorSheet`.

styles/elements.less
```diff
@@ -14,3 +14,7 @@ crucible-currency {
     text-align: center;
   }
 }
+
+[data-action] {
```
aaclayton: I worry about potential side effects of this rule, although it seems sensible. Out of some caution, let's scope this a bit more tightly to some selector on actor sheets specifically like `span.hex[data-action]` or similar.

aaclayton:
There's a few tweaks I want to make but I will merge this first and then make those in a follow-up commit. Thank you!

aaclayton (approved):

PR #534, opened 2025-12-30
[#515] Ensure action/item tabs are scrollable

### LINKED ISSUE #515: Correct some issues with window size and vertical scrolling on different sized monitors
I have an issue that I think it comes from my screen size (1920 * 1080 / window 1920 * 1017).

The talent points count in the talent tab makes a return to line.

Image

And the Target window of an effect is cut down if I'm not in full screen (even in VO)

Image

roth-michael:
Closes #515 
I wasn't able to reproduce the wrapping talent point count, but seems giving it `flex-wrap: nowrap` ought to cover it.

Notes:
- I don't love that now it's possible for there to de a double-scrollbar in the action description tab (one for the prosemirror element, one if the window is tiny) but that should be pretty rare.
- In theory, some _actor_ tabs (attributes & biography) could also be made scrollable. However, in my testing, it requires having a screen size small enough as to get the "Things won't work right" error notification to require that, so I've left that out for now.

### DIFF
--- a/module/applications/config/action-config.mjs
+++ b/module/applications/config/action-config.mjs
@@ -61,8 +61,7 @@ export default class CrucibleActionConfig extends api.HandlebarsApplicationMixin
     },
     description: {
       id: "description",
-      template: "systems/crucible/templates/sheets/action/description.hbs",
-      scrollable: [""]
+      template: "systems/crucible/templates/sheets/action/description.hbs"
     },
     usage: {
       id: "usage",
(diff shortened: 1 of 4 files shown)


module/applications/config/action-config.mjs
```diff
@@ -62,24 +62,29 @@ export default class CrucibleActionConfig extends api.HandlebarsApplicationMixin
     description: {
       id: "description",
       template: "systems/crucible/templates/sheets/action/description.hbs",
+      scrollable: [""]
```
aaclayton: I disagree with making the description tab scrollable. I agree that for other tabs, the tab itself should be scrollable, but I think scrolling on the description tab should come solely through the editor.

aaclayton (approved):

PR #539, opened 2025-12-30
[#521] Fix various talent config/automation issues

### LINKED ISSUE #521: Assorted talent configuration and automation issues
Some of these may be outdated and require reproduction:
 
- shield bash's link to staggered is broken in the mouseover
- Poisoner's Poison Blades action tries to attack yourself instead of applying the effect `poisonBlades`.
- In the new Spellcraft journal entry under game mechanics, the Surge rune has the same text as Pulse
- Illusion rune may be using the wrong training type for skill bonus
- When I use the beast shape action it doesn't pop up in the chat.
- Thread the needle is giving an error when used: ReferenceError: target is not defined. But i have a Target.
- Death Rune is now in Presence Tree but still lists intelligence as base (requirement is already set to presence)
- Mark Prey: Description says reduced damage, but the Action ist tagged as harmless and doesnt do any damage. Attacking a marked target (status effect) doesnt add the boons to the attack roll.
- shield charge is 0 actions? not sure if this is a bug or intentional and its meant to cost the actions for the strikes maybe instead?
- The cast a spell Action no longer works if you have 2 or fewer actions, but there is at least one spell that only requires 2 actions (touch)
- The place template button on the font of life ability minimizes all pages currently active, it also places a template but doesn't actually count the targets as targets. Same with "Rallying Cry"? managed to get that one to kind of work somehow on the third try but I'm not sure what made it work
- 0.8.1 and 0.8.3 Crucible, Double Punch costs -1A 1F, which gives you an Action point back upon use. It also strikes 3 times, contrary to the name and description.
- sometimes the guarded condition doesn't go away on turn start a round after using the defend action, not sure what specifically causes it but I don't think it happens every time as I remember it removing automatically a couple of times I think
- The aspect spell doesn't seem to provide its bonus to the life rune
- Guarded applies banes to attempts to heal a creature

roth-michael:
Closes #521 
One thing I did _not_ address was getting Aspect of Life/Soul to work. Not sure how we want to go about that, currently.
Changes that were _not_ simply pack data or hook fixes:
- The logic for the `movement` tag seemed to be assuming that _all_ such actions were actual "Move" actions. As such, cost was pretty much hardcoded to either 0 or 1, I think. Changed it so that `this.cost.action` is only overridden if it's an actual "Move" action, and otherwise it is reduced by 1 if there's a free movement available.
- Death scaling from Intellect -> Presence
- Round-based durations were lasting one round too long. `elapsed > rounds` -> `elapsed >= rounds`.

### DIFF
--- a/module/hooks/action.mjs
+++ b/module/hooks/action.mjs
@@ -566,7 +566,6 @@ HOOKS.poisonIngest = {
 
 HOOKS.rallyingCry = {
   postActivate(outcome) {
-    if ( outcome.self ) return;
     const amount = this.actor.abilities.presence.value;
     outcome.resources.morale = (outcome.resources.morale || 0) + amount;
   }
--- a/_source/talent/Rallying_Cry_rallyingcry00000.yml
+++ b/_source/talent/Rallying_Cry_rallyingcry00000.yml
@@ -29,7 +29,7 @@ system:
         size: 12
       effects:
         - name: Rallying Cry
-          scope: '2'
+          scope: '4'
           result:
             type: success
             all: false
@@ -67,8 +67,8 @@ _stats:
   systemId: crucible
   systemVersion: 0.8.3
   createdTime: 1766948760410
-  modifiedTime: 1767116962095
-  lastModifiedBy: ifaQA7jTJsr3UWFE
+  modifiedTime: 1766950680056
+  lastModifiedBy: AnoypGxxNIMOS0XY
 folder: kvXzGhvDAbUcQsXM
 ownership:
   default: 0


module/const/action.mjs
```diff
@@ -324,7 +324,8 @@ export const TAGS = {
     prepare() {
       const stride = this.actor.system.movement.stride;
       const movement = this.usage.movement || this.actor.getMovementActionCost(stride);
-      this.cost.action = movement.cost;
+      if ( this.id === "move" ) this.cost.action = movement.cost;
+      else if ( this.actor.system.hasFreeMove ) this.cost.action = Math.max(0, this.cost.action - 1);
```
aaclayton: Since `getMovementActionCost` is an Actor method, I think this should change to something like this where all the logic is placed into `getMovementActionCost` which is passed the `CrucibleAction` instance.
```js
const movement = this.actor.getMovementActionCost(this);
this.cost.action = movement.cost;
```

aaclayton (changes requested):
Good looking changes, thank you! Some iteration requests though.

module/documents/actor.mjs
```diff
@@ -1211,7 +1211,7 @@ export default class CrucibleActor extends Actor {
       const {startRound, rounds} = effect.duration;
       if ( Number.isNumeric(rounds) ) {
         const elapsed = game.combat.round - startRound;
-        if ( elapsed > rounds ) {
+        if ( elapsed >= rounds ) {
```
aaclayton: I don't think this is correct and seems unrelated to this PR. Context?

module/hooks/action.mjs
```diff
@@ -316,6 +316,28 @@ HOOKS.feintingStrike = {
 
 /* -------------------------------------------- */
 
+HOOKS.fontOfLife = {
+  postActivate(outcome) {
+    if ( outcome.self ) return;
+    const amount = this.actor.abilities.wisdom.value;
+    outcome.resources.health = (outcome.resources.health || 0) + amount;
+  },
+  confirm(reverse) {
```
aaclayton: Instead of pushing a new effect in `confirm` can we update a preconfigured `outcome.effects[0]` in `postActivate`?

module/documents/actor.mjs
roth-michael: >- sometimes the guarded condition doesn't go away on turn start a round after using the defend action, not sure what specifically causes it but I don't think it happens every time as I remember it removing automatically a couple of times I think

From #521 is the context. Example:
Round 2, a hero uses Defend.
Round 3, their turn starts. `game.combat.round` is 3. `startRound` is 2. So `elapsed` is 1. `rounds` is _also_ `1`, since that's what the duration is set to. `1 > 1` is `false`, so Defend won't wear off, even though this is indeed the start of their next turn.

module/hooks/action.mjs
roth-michael: Can do, yeah. Currently there's a self-scoped effect, but since that's purely informational without Auras might as well make that one ally-scoped and use it.

module/hooks/action.mjs
```diff
@@ -548,6 +566,7 @@ HOOKS.poisonIngest = {
 
 HOOKS.rallyingCry = {
   postActivate(outcome) {
+    if ( outcome.self ) return;
```
aaclayton: With my change to rallying cry being a heroism spender, I had envisioned that it would now apply to yourself as well.

PR #540, opened 2025-12-30
Adapt Archetype schema to allow for specifying level for talents

roth-michael:
So far, just modifying the schema and making any required code changes so that everything works as it already did. `uuid.item ?? uuid` is awful each time it comes up, rest assured it's a temporary measure, since everywhere it's used we'll be needing new explicit logic to handle the `level` aspect of things (in a follow-up PR).

### DIFF
--- a/module/applications/sheets/item-actor-details-sheet.mjs
+++ b/module/applications/sheets/item-actor-details-sheet.mjs
@@ -58,10 +58,12 @@ export default class CrucibleActorDetailsItemSheet extends CrucibleBaseItemSheet
    */
   async _prepareTalents() {
     const uuids = this.document.system.talents;
-    const promises = uuids.map(async (uuid) => {
-      const talent = (await fromUuid(uuid)) ?? await new Item.implementation({type: "talent", name: "INVALID"});
-      return talent.renderInline({showRemove: this.isEditable});
-    });
+    const promises = [];
+    for ( const uuid of uuids ) {
+      let talent = await fromUuid(uuid);
+      talent ||= new Item.implementation({type: "talent", name: "INVALID"});
+      promises.push(talent.renderInline({showRemove: this.isEditable}));
+    }
     return Promise.all(promises);
   }
 
--- a/module/applications/sheets/item-archetype-sheet.mjs
+++ b/module/applications/sheets/item-archetype-sheet.mjs
@@ -121,6 +121,7 @@ export default class CrucibleArchetypeItemSheet extends CrucibleBackgroundItemSh
 
   /* -------------------------------------------- */
 
+  // TODO 541
   /** @override */
   async _prepareTalents() {
     const talents = this.document.system.talents;


module/applications/sheets/item-actor-details-sheet.mjs
```diff
@@ -121,7 +121,10 @@ export default class CrucibleActorDetailsItemSheet extends CrucibleBaseItemSheet
   async #onDropTalent(event) {
     const data = CONFIG.ux.TextEditor.getDragEventData(event);
     const talents = this.document.system.talents;
-    if ( (data.type !== "Item") || talents.has(data.uuid) ) return;
+    const hasLeveledTalents = this.document.type === "archetype";
```
aaclayton: For the very short term, can we leave a `// TODO [issue number here]` anywhere that we have to logically condition on whether we have leveled talents or not so we can identify all the places that we need to circle back to later?

module/models/actor-adversary.mjs
```diff
@@ -364,7 +364,7 @@ export default class CrucibleAdversaryActor extends CrucibleBaseActor {
     /** @deprecated since 0.7.3 */
     if ( source.details?.archetype ) crucible.api.models.CrucibleArchetypeItem.migrateData(source.details.archetype);
     /** @deprecated since 0.7.3 */
-    if ( source.details?.taxonomy ) crucible.api.models.CrucibleArchetypeItem.migrateData(source.details.taxonomy);
+    if ( source.details?.taxonomy ) crucible.api.models.CrucibleTaxonomyItem.migrateData(source.details.taxonomy);
```
aaclayton: oops!

module/models/item-archetype.mjs
```diff
@@ -20,7 +20,10 @@ export default class CrucibleArchetypeItem extends foundry.abstract.TypeDataMode
         obj[ability.id] = new fields.NumberField({...nullableInteger, initial: 2, min: 0, max: 6})
         return obj;
       }, {}), {validate: CrucibleArchetypeItem.#validateAbilities}),
-      talents: new fields.SetField(new fields.DocumentUUIDField({type: "Item"})),
+      talents: new fields.ArrayField(new fields.SchemaField({
+        item: new fields.DocumentUUIDField({type: "Item"}),
+        level: new fields.NumberField({required: true, nullable: true, integer: true, initial: null})
+      })),
```
aaclayton: We could consider using a `TypedObjectField` here where the keys are the talent UUIDs and the values are the required levels (or maybe better a schema field if we anticipate other potential metadata). This would have the side effect of avoiding duplicates in the array since the UUID keys would be unique.
```suggestion
      talents: new fields.TypedObjectField(new fields.SchemaField({
        level: new fields.NumberField({required: true, nullable: true, integer: true, initial: null})
      }), {validateKey: itemUuidField.validate}),
```

module/applications/sheets/item-archetype-sheet.mjs
```diff
@@ -121,6 +121,26 @@ export default class CrucibleArchetypeItemSheet extends CrucibleBackgroundItemSh
...
+      const talent = await fromUuid(uuid);
+      if ( !talent ) return {uuid, name: "INVALID", img: "", description: "", tags: {}};
+      return {
+        uuid,
+        name: talent.name,
+        img: talent.img,
+        description: await CONFIG.ux.TextEditor.enrichHTML(talent.system.description),
+        tags: talent.getTags(),
+        item: await talent.renderInline({showRemove: this.isEditable})
+      }
```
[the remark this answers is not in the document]
roth-michael: Good point - cleaned this up both here and in `CrucibleActorDetailsItemSheet`.

aaclayton (approved):

PR #543, opened 2025-12-31
[#512] Clicking resource pips adjust the amount of that resource

### LINKED ISSUE #512: Support clicking on pips on the actor sheet to adjust resource values to the indexed number of the clicked-on pip as an alternative to typing in a number.

roth-michael:
Closes #512 
No super clean way to handle this for >6 action or >12 focus, but I'm happy with the current solution (if you click on a "doubled" pip, it'll offset the change accordingly - just means you can't _increase_ into a "doubled" state, only decrease from it).

### DIFF
--- a/module/applications/sheets/base-actor-sheet.mjs
+++ b/module/applications/sheets/base-actor-sheet.mjs
@@ -1039,10 +1039,11 @@ export default class CrucibleBaseActorSheet extends api.HandlebarsApplicationMix
     const {resource, index} = target.dataset;
     const maxPips = CrucibleBaseActorSheet.#maxResourcePips[resource];
     const offsetValue = Number(index) + 1 + ((event.button === 2) ? maxPips : 0);
-    let resourceValue = this.actor.resources[resource].value;
-    if ( resourceValue === offsetValue ) resourceValue--;
-    else resourceValue = offsetValue;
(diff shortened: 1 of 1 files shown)


module/applications/sheets/base-actor-sheet.mjs
```diff
@@ -27,7 +27,8 @@ export default class CrucibleBaseActorSheet extends api.HandlebarsApplicationMix
       effectDelete: CrucibleBaseActorSheet.#onEffectDelete,
       effectToggle: CrucibleBaseActorSheet.#onEffectToggle,
       expandSection: CrucibleBaseActorSheet.#onExpandSection,
-      skillRoll: CrucibleBaseActorSheet.#onSkillRoll
+      skillRoll: CrucibleBaseActorSheet.#onSkillRoll,
+      togglePip: CrucibleBaseActorSheet.#onTogglePip
```
aaclayton: Let's call this "resourcePip" to be a bit more clear.
```suggestion
      resourcePip: CrucibleBaseActorSheet.#onResourcePip
```

aaclayton (changes requested):
Needs a few changes as well as conflict resolution with some other changes in the `base-actor-sheet` file.

aaclayton (changes requested):

PR #544, opened 2026-01-01
[#453] Improve Action `effects` field definition, modify rendering accordingly

### LINKED ISSUE #453: Change action effect `scope` from a `StringField` to a `NumberField`
Carry-on from https://github.com/foundryvtt/crucible/pull/443

`scope` on an effect is currently stored as a string rather than a number, so self-scoped effects (set through the UI) on others-scoped actions won't ever register as needing to apply to self. Ideal solution is probably to change from a `StringField` to a `NumberField` in `CrucibleActionConfig.#defineEffectSchema`, but that won't handle existing items in the world, so instead just casting to a `Number` when doing the check

roth-michael:
Closes #453 
`effects` field in `CrucibleAction` schema now covers all the fields we reasonably support (besides, I guess, `flags` - but those would never be persisted anyway, only ever rarely set in a hook). As such, no more on-the-fly effect schema creation for rendering purposes. Since the inline template is no longer being programmatically accessed, I just moved it into the main `action/effects.hbs`. Since the schema isn't being created anew for each effect, also means we have to specify the `name` for each input, so I took inspiration there from how `changes` are handled on the AE Config sheet.
Also adjusted the `CrucibleAction##attachOutcomeEvents` logic to pull the system from the effect data (currently won't have anything, but means we don't need to assert its existence), and cleaned up areas where we had guardrails for improperly formatted effect data.

### DIFF
--- a/module/applications/config/action-config.mjs
+++ b/module/applications/config/action-config.mjs
@@ -165,13 +165,10 @@ export default class CrucibleActionConfig extends api.HandlebarsApplicationMixin
    */
   #prepareEffects() {
     const effects = this.action.toObject().effects;
-    return effects.map((e, i) => Object.assign(
-      e,
-      ["name", "scope", "result.type", "result.all", "statuses", "duration.turns", "duration.rounds"].reduce((paths, fieldName) => {
-        paths[`${fieldName.replaceAll(".", "")}Path`] = `effects.${i}.${fieldName}`;
-        return paths;
-      }, {})
-    ));
+    for ( const [i, effect] of effects.entries() ) {
+      effect.fieldPath = `effects.${i}`;
+    }
+    return effects;
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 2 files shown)


module/models/action.mjs
```diff
@@ -265,7 +268,22 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
+          all: new fields.BooleanField({initial: false})
+        }),
+        statuses: new fields.SetField(new fields.StringField({choices: statusEffectsObj})),
+        duration: new fields.SchemaField({
+          turns: new fields.NumberField({nullable: true, initial: null, integer: true, min: 0}),
+          rounds: new fields.NumberField({nullable: true, initial: null, integer: true, min: 0})
+        }),
+        system: new fields.SchemaField({
+          dot: crucible.api.models.CrucibleBaseActiveEffect.defineSchema().dot
+        })
```
aaclayton: I think we should allow the action `effects.element.system` to contain any/all of the fields that the `CrucibleBaseActiveEffect` supports.

PR #550, opened 2026-01-02
Ensure `CrucibleAction##applyOutcomeEffect` treats `success` more leniently, and `all` _less_ leniently

roth-michael:
Also updated the typedef for `ActionEffect`


aaclayton (approved):
Thanks!

PR #560, opened 2026-01-07
Move all nonstandard effect field usage to `postActivate`, ensure all spell component hooks are properly called

roth-michael:
Also deduplicated some code in the spellcraft hooks.

### DIFF
--- a/module/hooks/spellcraft.mjs
+++ b/module/hooks/spellcraft.mjs
@@ -2,6 +2,29 @@ const HOOKS = {};
 
 /* -------------------------------------------- */
 
+function prepareSummon(level) {
+  this.tags.add("summon");
+  this.usage.hasDice = false;
+  
+  // Configure summon data
+  const summonUUIDs = SYSTEM.SPELL.GESTURE_SUMMONS[this.gesture.id];
+  const actorUuid = summonUUIDs[this.rune.id] || summonUUIDs.fallback;
+  const tokenData = {
+    delta: {
+      system: {
+        details: {
+          level,
+          rank: "minion"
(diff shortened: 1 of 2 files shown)


module/hooks/spellcraft.mjs
```diff
@@ -76,53 +75,62 @@ HOOKS.conjure = {
...
       }
     };
+    const effectId = SYSTEM.EFFECTS.getEffectId(this.gesture.id);
     this.usage.summons = [{actorUuid, tokenData, effectId}];
+  },
+  postActivate(outcome) {
+    if ( !outcome.self ) return;
+    const effectId = outcome.summons[0].effectId;
+    outcome.effects.push({
+      _id: effectId,
```
aaclayton:
```suggestion
    outcome.effects.push({
      _id: outcome.summons[0].effectId,
```

module/models/action.mjs
```diff
@@ -1087,13 +1088,13 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
   #attachOutcomeEffects(outcome) {
     for ( const effectData of this.effects ) {
       if ( !this.#applyOutcomeEffect(outcome, effectData) ) continue;
-      const {_id, name, changes, duration, statuses: statusesArray, system} = effectData;
+      const {name, changes, duration, statuses: statusesArray, system} = effectData;
       const statuses = new Set(statusesArray);
```
aaclayton: I think some of this may need to change now that we have explicit schema on `action.effects`. For example `effectData.statuses` should already be a `Set` now according to the data schema.

PR #564, opened 2026-01-08
[#561] Change `organic` property to `natural` property on Armor

### LINKED ISSUE #561: Eliminate the `organic` property of armor in favor of `natural` which can be shared by both weapons and armor equipment.

roth-michael:
Closes #561 
No pack data needed to be rebuilt (besides the adventure data, which I leave to you), so was fairly straightforward.


module/models/item-armor.mjs
aaclayton: One requested addition: during armor data preparation, let's automatically set the `natural` property if the armor belongs to the `natural` category.

aaclayton (changes requested):

PR #567, opened 2026-01-09
Add leveled talents to schema for all actor details items

roth-michael:
All actor details items (Ancestry, Archetype, Background, Taxonomy) now have the capacity to store talents alongside what level they should be granted at. Currently, level isn't taken into consideration at all.
Next steps:
- UI treatment for specifying level
- Logic for granting only level-appropriate talents, re-applying details items on advancement


aaclayton (approved):
Looks good to me, I can see things getting tidy. Let's move on to standardizing the UI for all 4 item types next!

PR #571, opened 2026-01-11
Respect details-granted talent levels, refresh details-granted talents on advancement

roth-michael:


module/documents/actor.mjs
```diff
@@ -2319,6 +2320,7 @@ export default class CrucibleActor extends Actor {
       this.#updateSize(data, options);
       this.#updatePace(data, options);
       this.#applyResourceStatuses(data);
+      this.#refreshDetailTalents(data);
```
aaclayton: We typically use the term "refresh" for things like synchronous UI update code.
```suggestion
      this.#grantDetailTalents(data);
```

module/documents/actor.mjs
```diff
@@ -2505,6 +2507,32 @@ export default class CrucibleActor extends Actor {
...
   /* -------------------------------------------- */
 
+  /**
+   * Ensure Talents granted by details items (Ancestry and Background for Heroes, Archetype and Taxonomy for Adversaries),
+   * are present if they should be, and not present if they should not be.
+   */
+  async #refreshDetailTalents(data) {
+    if ( !("level" in (data.system?.advancement ?? {})) ) return;
+    const effectiveLevel = Math.max(0, data.system.advancement.level);
+    for ( const itemType of ["ancestry", "archetype", "background", "taxonomy"] ) {
```
aaclayton: Suggest parallelizing this instead of serial await inside the loop. Predefine `toCreate` and `toDelete` arrays, populate them asynchronously inside a loop, then `Promise.all()`

If you make this inner part a shared function like `#prepareGrantedDetailTalents(talentUuids)` that returns those `{toCreate, toDelete}` arrays, you could invoke that same helper function in `_applyDetailItem`.

aaclayton (changes requested):

aaclayton (approved):
Nice work, thanks!

PR #573, opened 2026-01-13
Ensure non-self-scoped effects only apply to self if actually targeted, and self-scoped effects rely on other outcomes if necessary to determine application

roth-michael:
### Changes
- Added `isTarget` boolean to `CrucibleActionOutcome`, `true` when the outcome's target has been explicitly targeted, `false` in the case that it's a `self` outcome but `self` should _not_ be a target of the effects or resource changes (besides whatever the cost of action use is)
- When acquiring targets from a template, only ignore self if the action is _not_ set to target self
- ALL-scoped effects will no longer apply to self, _unless_ self is an actual target of the action
- SELF-scoped effects, therefore, have more detailed logic when determining whether to apply:
  - `any` and `custom` result types behave as normal
  - `success`, `successCritical`, `failure`, and `failureCritical` behave as normal _if_ there are rolls on the self-outcome
  - The above instead defer to any of the _other_ outcomes if there are no rolls on the self-outcome. For an example, see the changes to the Grapple action listed below
- Due to changes to ALL-scoped effects, Grapple's "Grappled" effect now only applies to the target. A new "Grappling" effect, self-scoped, has been added with otherwise the same configuration as "Grappled." Because its result type is "Success," it will only apply if the non-self outcome has a successful roll.
- Healing/Rallying Elixir/Tonic hooks have change to exit early if `!outcome.isTarget`, avoiding accidental healing or DOT effect application to self when targeting another.

#### Original text:
Might be nice to have a more generic way of determining whether the `self` outcome exists because self was targeted, or just because of resource consumption. This feels satisfactory to me for now, though.

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -889,7 +889,7 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
     const {x, y} = template;
     const targetDispositions = this.#getTargetDispositions();
     const tokens = canvas.tokens.quadtree.getObjects(bounds, {collisionTest: ({t: token}) => {
-      if ( token.actor === this.actor ) return false;
+      if ( !this.target.self && (token.actor === this.actor) ) return false;
       if ( !targetDispositions.includes(token.document.disposition) ) return false;
       if ( token.document.hidden ) return false;
       const hit = token.getHitRectangle();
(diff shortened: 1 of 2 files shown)


module/hooks/action.mjs
```diff
@@ -346,7 +346,7 @@ HOOKS.healingElixir = {
 
 HOOKS.healingTonic = {
   postActivate(outcome) {
-    if ( this.outcomes.size > 1 && outcome.self ) {
+    if ( !outcome.isTarget ) {
       outcome.effects = [];
       return;
     }
```
aaclayton: Shouldn't the action effect configuration itself ensure that effects are only applied to a target of the action? Why isn't this just:
```js
if ( !outcome.isTarget ) return;
```

module/hooks/action.mjs
```diff
@@ -590,7 +590,7 @@ HOOKS.rallyingElixir = {
 
 HOOKS.rallyingTonic = {
   postActivate(outcome) {
-    if ( this.outcomes.size > 1 && outcome.self ) {
+    if ( !isTarget ) {
```
aaclayton:
```suggestion
    if ( !outcome.isTarget ) {
```
Same question as above though, why do we need to clear the effects array?

module/models/action.mjs
```diff
@@ -1128,6 +1134,9 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
    */
   #applyOutcomeEffect(outcome, effectData) {
 
+    // Assert actually targeted
+    if ( !outcome.isTarget ) return false;
```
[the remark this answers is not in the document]
roth-michael: Agh, didn't consider self-scoped effects. `ALL`-scoped is actually, I suppose, the issue here; that's what the effect is currently scoped to. My assumption was that it really meant "all targeted" - currently we don't have a scope which corresponds to "whoever I targeted, give them and only them the effect."

PR #575, opened 2026-01-13
[#574] Implement non-combat handling for Counterspell, currently usable only via enricher

### LINKED ISSUE #574: Enable a workflow for non-combat uses of Counterspell
From Discord:

> we need to figure out an approach that allows counter-spell mechanics to be used outside of combat encounters. We need to think about a framework for counter-spell as a tool used in exploration, similar to picking the lock of a door. 
> 
> I'm envisioning the overall solution might look something like this:
> 
> 1. Think about how the counterspell action could be generalized so that details of the magical effect being countered could be configured as part of something like action.usage
> 2. Implement a text enricher with some syntax like [[/counterspell {difficulty} {rune} {gesture} {inflection}]] that provides a UX entry point for configuring a non-combat counterspell

roth-michael:
Closes #574
Notes: 
- In a good number of spots I just went with my gut on certain implementations, happy to rework bits as needed (enricher format, for instance: `[[/counterspell rune=earth gesture=pulse dc=20]]`).
- Threw in some counterspell-related localization while I was at it.
- ~~I don't love my addition to `chat.less`, but without it the whispered enrichers can potentially get very ugly. I didn't notice any real difference just _removing_ the `height` and `line-height` on lines 168-169 of `journal.less`, for what it's worth, which would accomplish the same thing. But felt like adding specificity was a safer bet than removing existing css without your go-ahead.~~
- ~~A potential improvement might be passing actor ID along in the whispered version of the enricher. As-is, if some player owns 2 actors and gets the whisper for one while they have the _other_ selected, the enricher will either roll counterspell for the selected one or warn the user that it can't.~~

### DIFF
--- a/module/models/counterspell-action.mjs
+++ b/module/models/counterspell-action.mjs
@@ -58,7 +58,7 @@ export default class CrucibleCounterspellAction extends CrucibleSpellAction {
     if ( !wasSpell ) {
       const error = game.i18n.localize("SPELL.COUNTERSPELL.WARNINGS.BadTarget");
       if ( target ) target.error = error;
-      if (options.strict) throw new Error(error);
+      if ( options.strict ) throw new Error(error);
     }
     return targets;
   }
@@ -79,12 +79,11 @@ export default class CrucibleCounterspellAction extends CrucibleSpellAction {
   /* -------------------------------------------- */
 
   /** @override */
+  // TODO: Store target message ID on action instead of flag, obviating the need for this override
   async _prepareMessage(targets, {confirmed}={}) {
     const messageData = await super._prepareMessage(targets, {confirmed});
-    // TODO: Only use this.usage.targetAction
     const lastAction = this.usage.targetAction ?? ChatMessage.implementation.getLastAction();
     if ( !lastAction?.message ) return messageData;
-    // TODO: Store message ID on action instead of in flag
     foundry.utils.setProperty(messageData, "flags.crucible.targetMessageId", lastAction.message.id);
     return messageData;
   }


module/enrichers.mjs
```diff
@@ -303,6 +309,119 @@ async function onClickAward(event) {
...
+  }
+  const {rune="lightning", gesture="touch", inflection, dc="20"} = parsed;
+  const dataset = {rune, gesture, dc};
+  if ( inflection ) dataset.inflection = inflection;
+
+  // Return the enriched content tag
+  const tag = document.createElement("enriched-content");
+  tag.classList.add("counterspell");
+  Object.assign(tag.dataset, dataset);
+  if ( game.user.isGM ) {
```
aaclayton: This shouldn't be GM only. We can render the enricher HTML for everyone if they have visibility to the HTML that contains it.

module/enrichers.mjs
```diff
@@ -303,6 +309,119 @@ async function onClickAward(event) {
...
+  const dc = Number(dcString);
+  const actor = inferEnricherActor();
+  const counterspellAction = actor?.actions.counterspell;
+
+  // If inferred can counterspell, prompt
+  if ( !game.user.isGM || counterspellAction ) {
+    if ( !counterspellAction ) return ui.notifications.warn(game.i18n.format("SPELL.COUNTERSPELL.WARNINGS.NoTalent", {actor: actor?.name}));
+    const action = counterspellAction.clone({tags: Array.from(counterspellAction.tags).findSplice(t => t === "reaction", "noncombat")});
+    action.usage.dc = dc;
+    action.usage.targetAction = new crucible.api.models.CrucibleSpellAction({rune, gesture, inflection});
```
aaclayton: I have had second thoughts about my proposal for `targetAction` being a full `CrucibleAction` instance. `action.usage` is (intentionally) ephemeral data that isn't persisted as part of the chat message. 

We would want *what action are you responding to* to be data that could be persisted because we need to know that on the confirmation side and cannot reliably re-infer it. 

We can use `usage.targetAction` as an ephemeral construction for that, but I think we may need something like `targetAction` as a string that points to a chat message ID if we want to record this sort of reaction chain.

We can do this in a separate PR. I'll open an issue so we track the work.

module/enrichers.mjs
```diff
@@ -430,7 +549,7 @@ async function onClickHazard(event) {
 
   // Iterate over actor targets
   for ( const actorId of targets ) {
-    const actor = game.actors.get(actorId);
+    const actor = game.actors.get(actorId) ?? await fromUuid(actorId);
```
aaclayton: Why was this needed? We shouldn't ever have a target that isn't a world actor, right?

module/hooks/action.mjs
```diff
@@ -193,30 +193,49 @@ HOOKS.conjureArmament = {
...
+    this.usage.context.tags.rune = game.i18n.format("SPELL.COMPONENTS.RuneSpecific", {rune: this.rune?.name ?? none})
+    this.usage.context.tags.gesture = game.i18n.format("SPELL.COMPONENTS.GestureSpecific", {gesture: this.gesture?.name ?? none})
   },
   preActivate() {
     SYSTEM.ACTION.TAGS.spell.preActivate.call(this);
   },
   async roll(outcome) {
-    outcome.usage.defenseType = "willpower"; // Maybe changed later
-    const {gesture: usedGesture, rune: usedRune} = ChatMessage.implementation.getLastAction();
+    outcome.targetAction ??= ChatMessage.implementation.getLastAction();
```
aaclayton: I think we should be using `this.usage.targetAction` here which would be the rehydrated action to which we are responding.

PR #578, opened 2026-01-14
[#477] Ensure Iconic Spells automatically set up scaling according to their prerequisite Runes & Gestures

### LINKED ISSUE #477: Add a property/method/getter so that iconic spell actions automatically scale based on their parent talent's rune & gesture prerequisites

roth-michael:
Closes #477 
Two pieces:
1. Add `scaling` getter in `CrucibleSpellItem` which gets the array of abilities the spell should scale with based on required Runes & Gestures
2. Use that getter in `iconicSpell`'s `prepare` hook, setting `scaling` and `usage.bonuses.ability` accordingly

Also added missing Telecognition hooks so that it actually performs a skill attack instead of doing nothing.

### DIFF
--- a/module/hooks/action.mjs
+++ b/module/hooks/action.mjs
@@ -742,6 +742,17 @@ HOOKS.spellband = {
   }
 }
 
+/* -------------------------------------------- */
+
+HOOKS.telecognition = {
+  prepare() {
+    this.usage.hasDice = true;
+  },
+  async roll(outcome) {
+    const roll = await this.actor.skillAttack(this, outcome);
+    if ( roll ) outcome.rolls.push(roll);
+  }
+}
 
 /* -------------------------------------------- */
 


aaclayton (changes requested):
I think that maybe to solve this problem we need to think carefully about what type of `CrucibleAction` class an iconic spell should be.

Is it a `CrucibleAction`? Is it a `CrucibleSpellAction`? Do we need something like a `CrucibleIconicSpellAction`? 

My *hunch* is that we should find the right way for it to be a `CrucibleSpellAction` and both composed spells as well as iconic spells would flow through a shared framework for "this is a spell". 

Doing that could automatically solve some things like scaling, but it would also come with some additional problems. Currently a `CrucibleSpellAction` has one `rune` , not multiple, whereas an iconic spell might involve multiple components. 

I think this deserves a discussion.

roth-michael:
Closed in lieu of #666

PR #602, opened 2026-01-20
Fully clear `_state` when restarting creation, store name separately on restart for use during initialization

roth-michael:
`HeroCreationSheet##characterName` is `null` by default. It's set to whatever `this._state.name` is when reset, and back to `null` after being fetched for state initialization. If it's _not_ null, `this._state.name` isn't set during `_configureRenderOptions`, that way we don't follow up a refresh with a nonempty `_state`.


aaclayton (changes requested):
I wonder if there's a better way to do this. 
1. Initialize `#characterName` as the starting actor name
2. Re-record `#characterName` if we ever call `_reset()`
3. Prepare `_state.name = this.#characterName`

That seems a little cleaner to me, would it work?

PR #615, opened 2026-01-29
[#258] Implement mechanical consequences of 0-score Abilities

### LINKED ISSUE #258: Introduce special rules for adversaries which have zero in a particular Ability score which makes them behave in a mechanically different way.

roth-michael:
Closes #258
- Handle equipment restriction (0 strength) in `CrucibleAdversaryActor#usesEquipment`
  - Note that this doesn't actually block _possession_ of physical items, just equipping them. Dunno if we want to do that or not
- Handle restricted reaction usage (0 dexterity) in `reaction` tag's `canUse` hook.
- Handle 0 initiative roll (0 dexterity) in `CrucibleCombatant#getInitiativeRoll` by setting `rollData.incapacitated`
- Handle no heroism generation (0 wisdom) in `CrucibleAction##canGenerateHeroism`
- Handle resources changes in `CrucibleBaseActor##prepareFinalResources`:
  - 0 Toughness: `morale.max += health.max`, then `health.max = 0`
  - 0 Presence: `health.max += morale.max`, then `morale.max = 0`
  - 0 Wisdom: `heroism.max = 0`
  - 0 Intellect: `focus.max = 0`
- Handle restricted usage based on scaling in `CrucibleAction#_canUse`:
  - 0 Strength: If scaling & all scaling is physical, block
  - 0 Wisdom: If scaling & all scaling is mental, block
- Handle changes to `CrucibleAdversaryActor` status getters:
  - 0 Toughness: `isDead` always `false`, `isIncapacitated` is `true` if `isBroken`
  - 0 Presence: `isBroken` always `false`

Unrelated change: Fixed `ACTION.WarningCannotSpendAction` from saying `Paralyzed` to `Incapacitated`.

### DIFF
--- a/module/models/actor-base.mjs
+++ b/module/models/actor-base.mjs
@@ -856,8 +856,7 @@ export default class CrucibleBaseActor extends foundry.abstract.TypeDataModel {
 
     // Specific logic for 0-ability actors
     if ( !this.abilities.toughness.value && !this.abilities.presence.value ) {
-      resources.health.max = 0;
-      resources.morale.max = 0;
+      resources.health.max = resources.morale.max = 0;
     } else if ( !this.abilities.toughness.value ) {
       resources.morale.max += resources.health.max;
       resources.health.max = 0;


module/models/actor-base.mjs
```diff
@@ -854,6 +854,17 @@ export default class CrucibleBaseActor extends foundry.abstract.TypeDataModel {
     // Compute resource maximums
     for ( const r of Object.values(resources) ) r.max = Math.max(r.base + r.bonus, 0);
 
+    // Specific logic for 0-ability actors
+    if ( !this.abilities.toughness.value ) {
```
aaclayton: Unlikely to occur, but we should maybe handle the special case of both zero toughness AND zero presence first, in which case both pools should just be zero I think.

aaclayton (approved):
Nice work with this. I have one small request regarding the edge case of 0 toughness AND 0 presence, but otherwise this is good to merge!

PR #616, opened 2026-01-29
[#620] Allow passing tooltips to `crucibleTags`, infer if not there

### LINKED ISSUE #620: Add tooltips to tags wherever they're rendered

roth-michael:
Closes #620 
Can pass along as one of the properties in the `tags` object `tooltip`, which will be assumed to be localized already, and will add a tooltip to the tag span. If not provided, will check whether the key of the tag corresponds to one in `SYSTEM.ACTION.TAGS`, if so grabbing its tooltip.
In the future we'd probably want to manually pass tooltips for certain property tags (like `noisy` on armor), but for now it should catch a good chunk of "what's this mean again?" instances.
Also adjusted context prep in `CrucibleActionConfig` to be able to use `crucibleTags`.

Example 1, hovering on a tag within a locked action tooltip:
image

Example 2, hovering on a tag in the Talents tab:
image

(In both examples, the mouse is hovering the Finesse tag)

### DIFF
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -245,9 +245,8 @@ Hooks.once("init", async function() {
         if ( tag.cssClasses ) classes += ` ${foundry.utils.escapeHTML(tag.cssClasses)}`;
         const label = foundry.utils.escapeHTML(tag.label ?? tag);
         const styleString = tag.color ? ` style="--tag-color: ${tag.color.css}"` : "";
-        const tooltip = tag.tooltip ?? SYSTEM.ACTION.TAGS[id]?.tooltip
-        const tooltipString = tooltip ? ` aria-label="${tooltip}" data-tooltip` : "";
-        return `<span class="${classes}" data-tag="${id}"${styleString}${tooltipString}>${label}</span>`;
+        const tooltipString = tag.tooltip ? ` data-crucible-tooltip-text="${tag.tooltip}"` : "";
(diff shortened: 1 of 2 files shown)


crucible.mjs
```diff
@@ -245,7 +245,9 @@ Hooks.once("init", async function() {
         if ( tag.cssClasses ) classes += ` ${foundry.utils.escapeHTML(tag.cssClasses)}`;
         const label = foundry.utils.escapeHTML(tag.label ?? tag);
         const styleString = tag.color ? ` style="--tag-color: ${tag.color.css}"` : "";
-        return `<span class="${classes}" data-tag="${id}"${styleString}>${label}</span>`;
+        const tooltip = tag.tooltip ?? SYSTEM.ACTION.TAGS[id]?.tooltip
+        const tooltipString = tooltip ? ` aria-label="${tooltip}" data-tooltip` : "";
+        return `<span class="${classes}" data-tag="${id}"${styleString}${tooltipString}>${label}</span>`;
```
aaclayton: I'm wondering about explicitly rendering tooltip HTML for every tag everywhere a tag appears. It could result in quite a lot of background HTML that isn't actively engaged with. I'm generally a fan of deferred tooltip rendering where we can and would be interested in doing something like `data-crucible-tooltip="tag" data-crucible-tag="noisy"` or something like that, allowing us to defer rendering of the tooltip HTML until a hover event occurs.

aaclayton (approved):
Nice

PR #617, opened 2026-01-29
[#574, #589] Add query for requesting specific Counterspell use, deduplicate enricher target selection logic

### LINKED ISSUE #589: Ensure Actors added to Hazard or Award enricher popup are World Actors, & return their `id` instead of `uuid`

roth-michael:
Sort of a two-parter:
1. Add new `counterspellRequest` query which takes actor ID & "targetAction" config info and prompts the queried user (and use that in the counterspell enricher)
2. Pull out the "Select party members or arbitrary actors" logic which would've been used in Award, Counterspell, and Hazard enrichers into a helper function with parameters for customizing as needed. 

~~As part of (2), filtered out any embedded or compendium actors from the response, and returned their ID instead of UUID, which closes #589. We could try to do live validation of the inputs for that instead of just silently ignoring bad data, but not sure if that's worth it.~~
Update to the struck through: Modified to filter out only compendium actors, and always return UUID. This way we can still safely `fromUuidSync`, while leaving the door open to potentially targeting unlinked token actors.

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -381,7 +381,7 @@ async function onClickCounterspell(event) {
   const actor = inferEnricherActor();
 
   // If inferred can counterspell, prompt
-  if ( actor?.actions.counterspell ) return crucible.api.models.CrucibleCounterspellAction.prompt({actorUuid: actor.uuid, rune, gesture, inflection, dc});
+  if ( actor?.actions.counterspell ) return crucible.api.models.CrucibleCounterspellAction.prompt(actor, {rune, gesture, inflection, dc});
 
   // Prompt GM to pick among counterspellable party members
   const actors = crucible.party?.system.members.reduce((acc, {actor}) => {
@@ -504,9 +504,7 @@ async function onClickHazard(event) {
 
   // Select a target
   const actor = inferEnricherActor();
-  let targets;
-  if ( actor ) targets = new Set([actor]);
-  else targets = await chooseActorsDialog();
+  const targets = actor ? [actor] : (await chooseActorsDialog());
 
   // Iterate over actor targets
   for ( const actor of targets ) {
--- a/module/models/counterspell-action.mjs
+++ b/module/models/counterspell-action.mjs
@@ -95,11 +95,11 @@ export default class CrucibleCounterspellAction extends CrucibleSpellAction {
   /**
    * Handle a request to use a non-combat Counterspell action
    * @param {CrucibleActor|string} actor  An Actor or UUID of an Actor to perform the action
-   * @param {object} [options]            Additional options
-   * @param {string} [options.rune]       The rune used on the to-be-counterspelled action
-   * @param {string} [options.gesture]    The gesture used on the to-be-counterspelled action
+   * @param {object} options              Additional options
+   * @param {string} options.rune         The rune used on the to-be-counterspelled action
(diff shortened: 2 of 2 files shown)


crucible.mjs
```diff
@@ -259,6 +259,7 @@ Hooks.once("init", async function() {
 
   // Queries
   CONFIG.queries.rollSkillRequest = dice.StandardCheck.handle.bind(dice.StandardCheck);
+  CONFIG.queries.counterspellRequest = models.CrucibleCounterspellAction.promptCounterspell;
```
aaclayton: Minor code style: let's rename the queries into a verb-based format like `requestCounterspell` and `requestSkillCheck`

module/models/counterspell-action.mjs
```diff
@@ -87,4 +87,26 @@ export default class CrucibleCounterspellAction extends CrucibleSpellAction {
...
+
+  /**
+   * Handle a request to use a non-combat Counterspell action
+   * @param {string} actorId    The ID of the actor who should use Counterspell
+   * @param {string} rune       The rune used on the to-be-counterspelled action
+   * @param {string} gesture    The gesture used on the to-be-counterspelled action
+   * @param {string} inflection The inflection used on the to-be-counterspelled action
+   * @param {number} dc         The DC for the Counterspell action
+   */
+  static async promptCounterspell({actorId, rune, gesture, inflection, dc}={}) {
```
aaclayton: Perhaps "Counterspell" is implied here and we might just go with `static async prompt()`?

This is potentially an interface that might move upstream at some point to become a general action prompt framework?

module/models/counterspell-action.mjs
```diff
@@ -87,4 +87,26 @@ export default class CrucibleCounterspellAction extends CrucibleSpellAction {
...
+    const actor = game.actors.get(actorId);
+    const counterspellAction = actor?.actions.counterspell;
+    if ( !counterspellAction ) return;
+    const action = counterspellAction.clone({tags: Array.from(counterspellAction.tags).findSplice(t => t === "reaction", "noncombat")});
+    action.usage.dc = dc;
+    action.usage.targetAction = new crucible.api.models.CrucibleSpellAction({rune, gesture, inflection});
```
aaclayton: If this were to be part of a more general action-prompt system, we might want to consider the possibility of a specific target action rather than a generic one. Maybe out of scope of this PR, but worth considering.

module/enrichers.mjs
```diff
@@ -510,45 +506,13 @@ async function onClickHazard(event) {
...
-    });
-    const response = await foundry.applications.api.DialogV2.input({
-      window: {title: "Choose Target", icon: "fa-solid fa-bullseye"},
-      content: `\
-      ${partyMember.outerHTML}${anyActor.outerHTML}
-      `,
-    });
-    targets = new Set([...response.partyMember, ...response.anyActor]);
-  }
+  else targets = await chooseActorsDialog();
```
aaclayton: Hmm.... I'm wondering now whether `chooseActorsDialog` should simply return `Actor` instances.

It's not very nice putting the UUIDs from this result into a set and then iterating over them again to get Actors back in the subsequent loop. 

I think we should change it to return Actor instances and then we eliminate all the places we have to do things like
```js
targets.map(uuid => fromUuidSync(uuid))...
```

PR #618, opened 2026-02-01
Fix bug in `bodyBlock` hook

roth-michael:
Wonder if we should also handle the confirmation bit of this similarly to counterspell. Only difference would be that instead of just marking it negated, we might want to update the outcome retroactively to "blocked," which would be a more involved process.


aaclayton (approved):
Approved, regarding your comment I think this would end up using some common infrastructure for an intervention-style reaction that could potentially negate a prior action. We should have common framework for this for Counterspell and other actions, but that is a bigger chunk of work.

PR #621, opened 2026-02-02
[#604] Reload weapons & prompt to pick up dropped items at combat end

### LINKED ISSUE #604: Have `reload` weapons begin combat loaded, or auto-reload at the end of combat

roth-michael:
Closes #604 
Currently only prompts a user who has their `character` set to the actor.

Cleaned up some logic around combatant/combat deletion:
- On all clients:
  - When a combatant is removed from combat directly, reset & re-render (same as before, but removed an update which might've been called on the wrong clients)
  - When a combatant is removed from combat because combat is deleted, reset & re-render (was not resetting previously - honestly, not sure if it needs to be in either spot)
- On Active GM client, in addition to the above:
  - When a combatant is removed from combat directly, restore heroism to 0, remove delay flags, reset flanking
  - When a combatant is removed from combat because combat is deleted, do the same, but batch the heroism & delay flag updates for all combatants
  - In both cases, if a User is active who has their `character` set to the combatant's actor, and the actor is not incapacitated, and the actor has some dropped items, prompt User for whether to recover dropped items. If yes, attempt to recover. Then, if actor is not incapacitated, reload any equipped weapons in need of loading

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1316,6 +1316,7 @@ export default class CrucibleActor extends Actor {
     
     // Prompt designated user to pick up dropped items
     // TODO: Consider whether to extend this dialog to an Owner who does not have this actor as their character
+    const itemUpdates = [];
     const designatedUser = game.users.getDesignatedUser(user => {
       return user.active && (user.character === this);
     });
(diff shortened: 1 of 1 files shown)


module/documents/actor.mjs
```diff
@@ -1288,17 +1288,60 @@ export default class CrucibleActor extends Actor {
...
+  
+      // Clear flanking
+      if ( this.statuses.has("flanked") ) await this.commitFlanking();
+    }
+    
+    // Prompt designated user to pick up dropped items
+    // TODO: Consider whether to extend this dialog to an Owner who does not have this actor as their character
+    const designatedUser = game.users.getDesignatedUser(user => {
+      if ( !user.active ) return false;
+      return user.character === this;
```
aaclayton:
```suggestion
      return user.active && (user.character === this);
```

module/documents/actor.mjs
```diff
@@ -1288,17 +1288,60 @@ export default class CrucibleActor extends Actor {
...
+    });
+    const droppedItems = this.items.filter(i => i.system.dropped);
+    if ( designatedUser && !this.isIncapacitated && droppedItems.length ) {
```
aaclayton: If we don't have a `designatedUser` it would be good to ask the GM instead of asking nobody.

module/documents/actor.mjs
```diff
@@ -1288,17 +1288,66 @@ export default class CrucibleActor extends Actor {
...
+    const droppedItems = this.items.filter(i => i.system.dropped);
+    if ( !this.isIncapacitated && droppedItems.length ) {
+      const pickUp = await DialogV2.query(designatedUser ?? game.user, "confirm", {
+        window: { title: "ITEM.ACTIONS.RecoverAllTitle" },
+        content: game.i18n.format("ITEM.ACTIONS.RecoverAllContent", {actor: this.name})
+      });
+      if ( pickUp ) {
+        for ( const item of droppedItems ) {
+          try {
+            await this.equipItem(item);
```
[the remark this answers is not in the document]
roth-michael: Works for me, that was actually my original intent before I realized that re-equipping required sequential changes.

PR #625, opened 2026-02-02
[#623] Block non-composed-spell action use based on free hands (or spell hands, for spells)

### LINKED ISSUE #623: Properly block non-spell actions requiring more free hands than are available
Currently, we override `hands` cost to 0 in `CrucibleAction#_prepareData`, and don't account for it at all in `CrucibleAction#_canUse`, only ever in the `composed` hook.

roth-michael:
Closes #623

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -36,6 +36,7 @@ import CrucibleActionConfig from "../applications/config/action-config.mjs";
  * @property {DiceCheckBonuses} bonuses     Roll bonuses applied to this action
  * @property {ActionContext} context        Action usage context
  * @property {boolean} hasDice              Does this action involve the rolling a dice check?
+ * @property {number} [availableHands]      How many hands does the actor this action is on have available?
  * @property {string} [rollMode]            A dice roll mode to apply to the message
  * @property {string} [defenseType]         A special defense type being targeted
  * @property {string} [skillId]             A skill ID that is being used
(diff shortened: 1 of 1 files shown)


module/const/action.mjs
```diff
@@ -446,12 +446,6 @@ export const TAGS = {
         icon: "fa-solid fa-sparkles",
         tags: {}
       });
-    },
-    preActivate() {
-      let hands = this.cost.hands ?? 0;
-      if ( hands > this.actor.equipment.weapons.spellHands ) {
-        throw new Error(`The "${this.name}" spell requires ${this.cost.hands} free hands to cast.`);
-      }
```
roth-michael: Neither this nor the `composed` hook's `preActivate` are necessary now that `_canUse` will properly throw, so removed them.

module/models/action.mjs
```diff
@@ -1377,6 +1377,18 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
       }));
     }
 
+    // Cannot afford hands cost
+    const {freeHands, spellHands} = this.actor.equipment.weapons;
+    const plurals = new Intl.PluralRules(game.i18n.lang);
+    if ( this.cost.hands > (this.tags.has("spell") ? spellHands : freeHands) ) {
```
roth-michael: Figure `spell` tag means `spellHands`, anything else means `freeHands`. If ever there are exceptions to this, we can add them here.

PR #640, opened 2026-02-05
[#636] `crawl`-action movement no longer removes `prone` status

### LINKED ISSUE #636: Ensure movement with the "Crawl" movement-action does not automatically remove Prone

roth-michael:
Closes #636 
Not sure if there's a better way to say "did any of this movement not use crawl" but I think this works.

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -352,8 +352,10 @@ export const TAGS = {
 
       // Remove prone condition upon movement confirmation if non-crawl action
       if ( this.actor.statuses.has("prone") ) {
-        const movementPath = this.token?.movementHistory.filter(m => m.movementId === self.movement) ?? [];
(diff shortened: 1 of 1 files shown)


module/const/action.mjs
```diff
@@ -339,20 +339,23 @@ export const TAGS = {
...
         }
         return;
       }
 
-      // Remove prone condition upon movement confirmation
+      // Remove prone condition upon movement confirmation if non-crawl action
       if ( this.actor.statuses.has("prone") ) {
-        await this.actor.toggleStatusEffect("prone", {active: false});
+        const movementHistory = this.token?.movementHistory ?? [];
+        if ( !movementHistory.every(m => (m.movementId !== self.movement) || (m.action === "crawl")) ) {
```
aaclayton: Could be micro-optimized to iterate in reverse order and know when to stop looking (after the movement ID no longer matches), but I think this is fine.

PR #651, opened 2026-02-06
[#645] Add `summons` to action schema & fieldset to action config

### LINKED ISSUE #645: For actions which have the `summon` tag, add a conditional fieldset to the ActionConfig sheet which allows configuration of which Actor UUID is summoned to enable UI-only configuration of "basic summons" without the need for an action hook.

roth-michael:
Closes #645 
As mentioned in the TODO, wonder what more makes sense to include in the UI currently/in the future. Part of me also wonders about dynamically hiding some fields for a summon action, and mandating "Summon" target type, but future considerations.

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -557,10 +557,7 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
 
     // Prepare Summons
     if ( this.summon.actorUuid ) {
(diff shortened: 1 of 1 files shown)


aaclayton (changes requested):
Couple small requests

aaclayton (approved):

PR #657, opened 2026-02-07
[#656] Show transfer effects on actor sheet, add suppression rules

### LINKED ISSUE #656: Display ActiveEffects that are transferred from an owned Item on the Effects tab of the character sheet.

roth-michael:
Closes #656 
Currently, they share the existing sections, but are differentiated from directly-embedded effects by the addition of a tag with the name of the item they're on. 
I did add one "Suppressed Effects" section, which supersedes others in priority (a suppressed effect will show there, regardless of whether it's disabled or not).
I've also added an `CrucibleActiveEffect#isSuppressed` getter override, which currently just checks a new `CrucibleItem#areEffectsSuppressed` getter if its parent is indeed a `CrucibleItem`.
That new getter, in turn, delegates to item subtypes. Currently the only one that makes sense to have its own logic, I think, is `CruciblePhysicalItem`, which always returns `false` (not suppressed) if the item isn't actually equipable (so, just schematics for now, I think), otherwise returns according to whether it's equipped and whether it's invested (if it requires investment).

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -27,7 +27,7 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
   /** @inheritDoc */
   get isSuppressed() {
     if ( super.isSuppressed ) return true;
-    if ( (this.parent instanceof crucible.api.documents.CrucibleItem) && this.parent.areEffectsSuppressed ) return true;
+    if ( (this.parent instanceof crucible.api.documents.CrucibleItem) && this.parent.activeEffectsSuppressed ) return true;
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


module/documents/item.mjs
```diff
@@ -39,6 +39,15 @@ export default class CrucibleItem extends foundry.documents.Item {
     return this.system.currentRank;
   }
 
+  /**
+   * Should this item's effects be suppressed?
+   * @type {boolean}
+   */
+  get areEffectsSuppressed() {
+    if ( this.system.areEffectsSuppressed ) return true;
```
aaclayton:
```suggestion
    return this.system.activeEffectsSuppressed || false;
```

module/documents/active-effect.mjs
```diff
@@ -24,6 +24,15 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  get isSuppressed() {
+    if ( super.isSuppressed ) return true;
+    if ( (this.parent instanceof crucible.api.documents.CrucibleItem) && this.parent.areEffectsSuppressed ) return true;
```
aaclayton:
```suggestion
    if ( (this.parent instanceof crucible.api.documents.CrucibleItem) && this.parent.activeEffectsSuppressed ) return true;
```

aaclayton (approved):

PR #665, opened 2026-02-22
Only apply prototype token defaults on actor creation if data isn't provided

roth-michael:
For actors created via the UI, will still define these defaults. Now an actor created programmatically (for instance, via Adventure import) that _has_ values already won't have them clobbered.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -2314,9 +2314,7 @@ export default class CrucibleActor extends Actor {
         Object.assign(prototypeTokenDefaults, {"sight.enabled": true, actorLink: true, disposition: 1});
         break;
       case "adversary":
-        prototypeTokenDefaults["sight.enabled"] = false;
-        prototypeTokenDefaults.actorLink = false;
-        prototypeTokenDefaults.disposition = -1;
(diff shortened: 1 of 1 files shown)


module/documents/actor.mjs
```diff
@@ -2308,15 +2308,23 @@ export default class CrucibleActor extends Actor {
...
 
     // Automatic Prototype Token configuration
-    updates.prototypeToken = {bar1: {attribute: "resources.health"}, bar2: {attribute: "resources.morale"}};
+    const prototypeTokenDefaults = {"bar1.attribute": "resources.health", "bar2.attribute": "resources.morale"};
     switch ( data.type ) {
       case "hero":
-        Object.assign(updates.prototypeToken, {sight: {enabled: true}, actorLink: true, disposition: 1});
+        prototypeTokenDefaults["sight.enabled"] = true;
+        prototypeTokenDefaults.actorLink = true;
+        prototypeTokenDefaults.disposition = 1;
```
aaclayton:
```suggestion
        Object.assign(prototypeTokenDefaults, {"sight.enabled": true, actorLink: true, disposition: 1});
```

aaclayton (approved):
Approved with changes

Issue #97, opened 2025-06-10
Implement support for a queried group check as a prerequisite for social and exploration challenge play


aaclayton:

Plushtoast:
If you could phrase out the general idea and specification I would deal with this topic.

PR #697, opened 2026-03-01
Add user feedback for party addition to request roll

Plushtoast:
Related to #97 

User: Clcky buuuuuttoooon no worky. Can eat buuuuuttoooon? tasty?
Button: *silence*
User: *Crunch*

Fix:

User: Clcky buuuuuttoooon no worky. Can eat buuuuuttoooon? tasty?
Button: Sorry pal no fun without a party
User: oooh

Possible further enhancement: could use game.users.map(x => x.character) as fallback party


Plushtoast:
I am not sure if member.actor can be null due to deleted actors (like the every returning better check token.actor)
If that is the case then i think the second loop is required. if it never can be null then yes

PR #699, opened 2026-03-01
Implement support for a queried group check as a prerequisite for soc…

### LINKED ISSUE #97: Implement support for a queried group check as a prerequisite for social and exploration challenge play

Plushtoast:
Ok here is a revised pull request for the group check feature.

I extracted the whole group check logic into its own class, so it is easier to review.

I mainly removed:
* additional entry points e.g. journal enricher and skill picker instant check application (will pr that later if this is accepted
* minor things

What is still there that is not api is the rendering of the corresponding chat card. I felt like I need that for better testing.
In the corresponding ticket you proposed to keep an dialog open which is updating once results come in.
Instead i chose a chat card which is fire and forget (dialogs not in the way) and has a persistent storage for the data.
It seems to be also a more common pattern among systems to control this via chat buttons

Currently the request is still the same like before from within the StandardCheckDialog.

On Request it creates this in the chat:

grafik

All actor owners receive a roll request dialog.
Actors without an owner have to be rolled by the gm

After each roll the card is updated

grafik

The buttons next to the actor provide these functions.

* Request roll again (e.g. due to disconnect)
* Roll on behalf (e.g. player was beamed up by an alien and alien does not know how to click roll)
* Skip actor (e.g. someone else failed so hard that there is no hope anymore)

If you still want me to remove the chat implementation, I can do that as well. But i think from here on the PR is less confusing.

Fixes #97


module/dice/standard-check-dialog.mjs
```diff
@@ -359,40 +368,8 @@ export default class StandardCheckDialog extends DialogV2 {
    * @this {StandardCheckDialog}
    */
   static async #requestSubmit(_event, _target) {
```
aaclayton: Maybe this should become a `GroupCheck` instance method somehow? Is `this.roll` a `GroupCheck` instance in this case? If so it could be:

```suggestion
  static async #requestSubmit(_event, _target) {
    await this.roll.requestSubmit();
    await this.close();
  }
```

module/dice/group-check.mjs
```diff
@@ -0,0 +1,538 @@
...
+  static CONFIG = Object.freeze({
+    /** @enum {string} Possible status values for an actor in a group check */
+    ACTOR_STATUS: Object.freeze({
+      PENDING: "pending",
+      COMPLETE: "complete",
+      ABORTED: "aborted",
+      SKIPPED: "skipped"
+    }),
+    /** @type {string} Key used for ChatMessage flags */
+    FLAG_KEY: "groupCheck"
```
aaclayton: Maybe this is also just a separate static. I'd recommend splitting these up and just doing:

```js
static FLAG_KEY;

static #STATUSES;
```

aaclayton (changes requested):
I still need to finish reviewing the `group-check.mjs` file, but I have to go AFK now for a couple hours so I'm posting the comments that I have so far and I'll finish the review later.

module/dice/group-check.mjs
```diff
@@ -0,0 +1,538 @@
...
+   * Derives status booleans, CSS classes, outcome/status labels, and roll context.
+   * @param {GroupCheckActorEntry} entry - The raw actor entry from flags
+   * @returns {object} The enriched entry for the Handlebars template
+   */
+  static #prepareActorTemplateData(entry) {
+    const { ACTOR_STATUS } = this.CONFIG;
+    const isComplete = entry.status === ACTOR_STATUS.COMPLETE;
+    const isPending = entry.status === ACTOR_STATUS.PENDING;
+    const isAborted = entry.status === ACTOR_STATUS.ABORTED;
+    const isSkipped = entry.status === ACTOR_STATUS.SKIPPED;
```
aaclayton: Instead of preparing these booleans and doing logic in the template, it would be better to prepare some data for status icon in javascript and render that icon directly in the template.

module/dice/group-check.mjs
```diff
@@ -0,0 +1,538 @@
...
+ * Creates a GM-whispered chat card that tracks individual roll statuses,
+ * dispatches roll queries to players, and finalizes results.
+ * @extends {StandardCheck}
+ */
+export default class GroupCheck extends StandardCheck {
+  /**
+ * The template path used to render group check chat cards.
+ * @type {string}
+ */
+  static GROUP_CHECK_TEMPLATE = "systems/crucible/templates/dice/group-check-chat.hbs";
```
aaclayton: This is only used inside the class, so perhaps should be private.

aaclayton:
> do you want to have the interaction via a chatcard or should i bringt that back into the dialog. the chat card stores the data in flags, which is basically extra data stored that might not be required anymore. if that would be a dialog the data could be discarded once the dialog is complete and you just have a plain chat card

I'm happy with the chat card proposal.

> do you need any sort of group result. currently it just basically list indiviudal results. but it could additionally do (if you define rules): define a threshold (e.g. 2 players need a success, a crit counts as 2 successes, a botch counts as -1 success), define an opposing role (higher role wins, define a defender (group) and attacker (group) higher sum wins, or define rolls with intervals e.g. a task needs 5 successes each roll takes 1h ingame, if the group does not achieve all 5 successes in one roll -- repeat duration increases ... etc

Yes, but let's keep that out of scope for this PR. I'd like to address the feedback here and get it merged and then we can take a next step to determine the aggregate group success or failure.

> some sort of individual boons pre selected for single characters (instead of everyone the same)

We will need to make sure this integrates in with the actor hooks system for skill checks, but this should happen automatically if you are careful to go through `Actor#getSkillCheck()`. Is this currently used?

> hidden DC

Can be out of scope for this PR.

Plushtoast:
Ok the PR is now targeting v14 branch

module/dice/group-check.mjs
```diff
@@ -0,0 +1,600 @@
...
+
+    const content = await this.#renderGroupCheckCard(groupCheckFlags);
+
+    const gmUserIds = game.users.filter(u => u.isGM).map(u => u.id);
+    const message = await ChatMessage.create({
+      content,
+      whisper: gmUserIds,
+      speaker: ChatMessage.getSpeaker({alias: _loc("DICE.GROUP_CHECK.Title")}),
+      flags: {crucible: {[GroupCheck.FLAG_KEY]: groupCheckFlags }}
+    });
```
roth-michael:
```suggestion
    const message = await ChatMessage.implementation.create({
      content,
      whisper: game.users.filter(u => u.isGM).map(u => u.id),
      speaker: ChatMessage.implementation.getSpeaker({alias: _loc("DICE.GROUP_CHECK.Title")}),
      flags: {crucible: {[GroupCheck.FLAG_KEY]: groupCheckFlags }}
    });
```
Could get fancy and make the filter -> map into `game.users.reduce((acc, u) => u.isGM ? [...acc, u.id] : acc, [])` but not sure it's worth the minor degradation of readability.

module/dice/group-check.mjs
```diff
@@ -0,0 +1,600 @@
...
+    const queued = previous.catch(() => {}).then(async () => {
+      const current = game.messages.get(message.id) ?? message;
+      const flags = foundry.utils.deepClone(current.flags.crucible?.[this.FLAG_KEY]);
+      if ( !flags ) return false;
+
+      const result = await mutator(flags, current);
+      if ( result === false ) return false;
+
+      const content = await this.#fromFlags(flags).#renderGroupCheckCard(flags);
+      const updateData = (result && (typeof result === "object") && !Array.isArray(result)) ? result : {};
```
roth-michael:
```suggestion
      const updateData = foundry.utils.isPlainObject(result) ? result : {};
```

roth-michael (changes requested):
Looks pretty good overall, and I think it's a huge UX improvement. The only things I've called out are style, docs, simplifications & stuff I maybe don't get - didn't find any bugs.

module/dice/group-check.mjs
```diff
@@ -0,0 +1,600 @@
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Apply the group DC to serialized roll data before it is stored by the GM.
+   * @param {object} rollData - Serialized roll data returned from a player or local roll
+   * @param {number} dc - The group check DC
+   * @returns {object} Normalized serialized roll data
+   */
+  static #applyGroupCheckDC(rollData, dc) {
```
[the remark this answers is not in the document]
roth-michael: Where it's used in `#dispatchGroupCheckQuery`, you can just set `response.data.dc = flags.dc` and pass that along, as far as I can tell. Similarly I think could do it in `#rollOnBehalfGroupCheck`. All the crucible-specific result stuff like `isSuccess`, `isFailure` etc are getters, not stored in the data itself. So I don't think you need to be building & then dumping the roll, as long as you're operating with data and not a `StandardCheck` itself (though even then it might work - it just feels not great).

module/dice/standard-check.mjs
```diff
@@ -83,6 +98,22 @@ export default class StandardCheck extends Roll {
...
+   */
+  static DEFENSE_TYPE = "DC";
+
+  /* -------------------------------------------- */
+
+  /**
+   * Default timeout for remote roll requests.
+   * @type {number}
+   */
+  static QUERY_TIMEOUT = 120_000;
```
roth-michael, line 113: I don't think we use `_` in numbers elsewhere - then again, I'm not sure if we have any numbers greater than 3 digits elsewhere

lang/en.json
```diff
@@ -1266,6 +1266,21 @@
...
+      "Status": {
+        "pending": "Pending",
+        "aborted": "Aborted",
+        "skipped": "Skipped"
+      },
+      "Resend": "Resend Request",
+      "RollOnBehalf": "Roll on Behalf",
+      "Skip": "Skip Actor",
+      "Finalize": "Reveal to Players"
+    },
```
roth-michael, line 1283: Minor enough not to block approval for me, but recently did some reorganization of i18n. Namespaces should be SCREAMING_SNAKE_CASE, keys (generally) PascalCase (but in the case of the statuses I think camelCase is fine - otherwise would need to change the enum to be `Pending`, `Aborted`, and `Skipped`), and sorting would go alphabetical namespaces -> alphabetical keys.
So in this case:

```json
    "GROUP_CHECK": {
      "STATUS": {
        "pending": "Pending",
        "aborted": "Aborted",
        "skipped": "Skipped"
      },
      "ChatTitle": "{skill} Check (DC {dc})",
      "DialogTitle": "{skill} Check: {name}","Title": "Group Check",
      "Finalize": "Reveal to Players",
      "NoUserForActor": "No active user found for: {name}",
      "Resend": "Resend Request",
      "RollOnBehalf": "Roll on Behalf",
      "Skip": "Skip Actor"
    },
```

(and then changing references from `GROUP_CHECK.Status.` to `GROUP_CHECK.STATUS.`)

roth-michael (approved):
Couple minor notes, both purely style.

module/dice/standard-check.mjs
Plushtoast: it's pretty readable

lang/en.json
Plushtoast: maybe a small script that resorts and warns on key that dont follow the rules would be nice. unless there is sth like esbuild for json as well (i dont know of such a thing)

roth-michael: Yeah that'd be super useful. I’m not sure if it's feasible because certain fields (like the literal `FIELDS` ones for document subtypes) at times need to _not_ follow that pattern. I’ll see though. Could save some minor headaches.

Plushtoast: could be at least a script in the node build thing. i for sure wouldnt review such typing and any new one trying to contribute (as myself) would have a far too long learning loop on a minor thing

aaclayton:
Rebased, resolved conflicts, made some review adjustments and merged via https://github.com/foundryvtt/crucible/pull/770

aaclayton:
Thank you @Plushtoast for the big contribution!

PR #700, opened 2026-03-03
[#702] Rework weapon choice framework to fix a number of bugs/bothers

### LINKED ISSUE #701: Miscellaneous 0.8.9 Bugfixes
- Composed spells sometimes target the wrong (previous spell's) defense
- Automatic adding of Damage Over Time from an effect with, for instance, the `poisoned` status on it, are not always being added properly
- `generic`-tagged actions are not displaying/adding roll to outcome unless it is >= a glancing success
- `generic`-tagged actions do not take multiplier, base, bonus into account
- Skill attacks erroneously apply skill bonus to base damage
- Damage Over Time doesn't respect Resolute or Invulnerable

### LINKED ISSUE #702: Fix a number of issues surrounding actions which don't consider multiple equipped weapons
- If your mainhand weapon costs more action than your offhand, Strike might disallow its use if you only have enough actions for the latter
- Using something like "Offhand Strike" with a reloadable ranged weapon does not cause it to become unloaded
- `natural`-tagged actions automatically fail for Hero actors
- `reload`-tagged actions don't let you pick which weapon you want to reload, if wielding two unloaded weapons
- Choosing a different weapon, when given the choice in the action use dialog, will not update ability/enchantment/skill fields

roth-michael:
Closes #702, Closes #701
Changes:
- When `usage.weapon` is set in `prepare`, it will only be set to a weapon which would cause `_canUse` to throw if there are no valid options
- Moved the setting of `system.loaded` to a `strike` hook instead of `ranged`, as some action (like Offhand Strike) would otherwise not be updating reloadable weapons' loaded status
- When doing a `natural`-tagged action, if multiple natural weapons are equipped, can choose which to attack with
- When using a `reload`-tagged action, if multiple unloaded weapons are equipped, can choose which to reload
- Choosing a different weapon in the action use dialog will now update the display of ability/enchantment/skill
- Also fixed `#prepareBaseResources` so that AEs can target `bonus` again

There's definitely room for further refinement - right now one can select, for instance, a Talisman to perform a ranged strike with, or a hand crossbow to try to Refocus with (the latter will yell, and it's better than before where sometimes you couldn't refocus at all, but still).

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -689,7 +689,7 @@ export const TAGS = {
     prepare() {
       this.usage.actorStatus.rangedAttack = true;
       if ( !this.usage.weapon ) {
-        const valid = this.getValidWeaponChoices(true);
+        const valid = this.getValidWeaponChoices({strict: true});
         valid.sort((a, b) => a.item.system.actionCost - b.item.system.actionCost);
         this.usage.weapon = valid[0]?.item;
       }
@@ -850,9 +850,9 @@ export const TAGS = {
       if ( r >= AttackRoll.RESULT_TYPES.GLANCE ) {
         roll.data.damage = {
           overflow: roll.overflow,
-          multiplier: bonuses.multiplier,
+          multiplier: bonuses.multiplier ?? 1,
           base: bonuses.base ?? 0,
-          bonus: bonuses.damageBonus,
+          bonus: bonuses.damageBonus ?? 0,
           resistance: target.getResistance(resource, damageType),
           type: damageType,
           resource: resource,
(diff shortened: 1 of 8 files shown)


module/const/action.mjs
```diff
@@ -850,9 +850,9 @@ export const TAGS = {
       if ( r >= AttackRoll.RESULT_TYPES.GLANCE ) {
         roll.data.damage = {
           overflow: roll.overflow,
-          multiplier: 1,
-          base: 0,
-          bonus: 0,
+          multiplier: bonuses.multiplier,
```
aaclayton: Let's include explicit fallbacks, `bonuses.multiplier ?? 1`, `bonuses.damageBonus ?? 0`, etc...

aaclayton (changes requested):
Some minor changes to request, overall looks good.

aaclayton (approved):
Thanks for the changes!

Issue #609, opened 2026-01-23
V14 Measured Template -> Scene Region Conversion


aaclayton:
- Replaced all `MeasuredTemplate` usage with usage of `RegionDocument`
- Replaced the previous ad-hoc template preview workflow with `canvas.regions.placeRegion` API
- Updated configuration of `SYSTEM.ACTION.TARGET_TYPES` with the addition of a default `ephemeral` boolean for whether the target type is typically persisted or not.
- Implement `onMove` constraints as part of region placement UX to require the region to be snapped to origin or placed within maximum range.
- Overhauled `CrucibleAction#region` to be a reference to a `RegionDocument` instance (potentially ephemeral or otherwise persited)
- `CrucibleAction#acquireTargets` replaced previous ad-hoc approach with usage of `RegionDocument#tokens`
- Persisted regions are eagerly created on action use, but only visible to the acting user and to Gamemasters. When the action is confirmed, the persisted region is revealed to all players.
- Add elevation ranges for created regions
- Add wall constraints for created regions
- Other miscellaneous tidying

PR #710, opened 2026-03-08
V14 regions

aaclayton:
Closes https://github.com/foundryvtt/crucible/issues/609
Closes https://github.com/foundryvtt/crucible/issues/712

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -13,7 +13,7 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
   /** @inheritDoc */
   async _onDelete(options, userId) {
     await super._onDelete(options, userId);
-    if ( game.user !== game.users.activeGM ) return;
+    if ( !game.user.isActiveGM ) return;
     for ( const uuid of this.system.regions ) {
       const region = await fromUuid(uuid);
       if ( region ) await region.delete();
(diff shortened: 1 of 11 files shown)


module/documents/active-effect.mjs
```diff
@@ -10,6 +10,18 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _onDelete(options, userId) {
+    await super._onDelete(options, userId);
+    if ( game.user !== game.users.activeGM ) return;
```
roth-michael: Any reason to use this over `!game.user.isActiveGM`?

roth-michael (commented):
Cursory pass, these were the only two things that occurred to me

module/models/action.mjs
```diff
@@ -869,48 +870,48 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
-    // Convert to target data structure
-    const targets = Array.from(tokens).map(CrucibleAction.#getTargetFromToken);
+    // Identify tokens contained within the region which match the correct disposition and visibility
+    const targets = [];
+    for ( const token of potentialTokens ) {
+      const tokenDoc = token.document;
+      if ( !this.target.self && (tokenDoc.actor === this.actor) ) continue;   // Exclude self
+      if ( !targetDispositions.includes(tokenDoc.disposition) ) continue;     // Require correct disposition
+      if ( tokenDoc.hidden ) continue;                                        // Ignore hidden
+      if ( !tokenDoc.testInsideRegion(this.region) ) continue;                // Require region containment
```
dev7355608:
```suggestion
      if ( !tokenDoc.testInsideRegion(this.region, tokenDoc._source) ) continue;                // Require region containment
```

Otherwise we test in the current animation state of the token.

The token is quadtree entry is never in an animated state.

PR #736, opened 2026-03-13
Fix missing i18n and localize all user-facing errors/warnings

### LINKED ISSUE #516: [Localization] Chat / Combat tracker / Use action / Notification
Some things I discovered
Combat tracker
Image

Chat
Image

Image

Image

Use action (recover something droped, and choose weapon for a strike)
Image

Image

Notifications (maybe there is others I didn't see)
No target for a strike
Image
Action that must be after a strike
Image
Not able to use a reaction
Image
Not able in combat
Image
Not enough uses of a consumable
Image

roth-michael:
Closes #516 
The only _change_ in what an english-using user will see is that the "must be a GM" warnings for awarding & revoking milestones now just use the "must be a GM" award enricher text. This is because it's impossible for those dialogs to be opened any other way (we exit early without feedback if a non-gm clicks the milestone award button on a group sheet), and so I didn't think it made sense to create separate localizations for them.

As far as I am aware, the only remaining non-localized errors are ones which will only show up either in the console or as a result of something which can only be done programmatically.

I changed calls of `ui.notifications.warn("Something", {localize: true})` to `ui.notifications.warn(_loc("Something"))` for brevity and to make it more easily searchable where we're using localization, or for simplicity if we wanted to set up some sort of extension like i18n-ally.

**TBD**: Should the caught errors in `CrucibleAction##use` be logged in their entirety as they are now, or only their message? In their entirety means the forced `Error: ` prefix, but also means the stack automatically gets logged to console.

### DIFF
--- a/module/applications/sheets/item-actor-details-sheet.mjs
+++ b/module/applications/sheets/item-actor-details-sheet.mjs
@@ -139,11 +139,11 @@ export default class CrucibleActorDetailsItemSheet extends CrucibleBaseItemSheet
     if ( talents.some(({item}) => item === data.uuid) ) return;
     const talent = await fromUuid(data.uuid);
     if ( talent?.type !== "talent" ) {
-      ui.notifications.warn("ITEM.WARNINGS.NotTalent", {localize: true});
+      ui.notifications.warn(_loc("ITEM.WARNINGS.NotTalent"));
       return;
     }
     if ( talent.system.node?.tier && (talent.system.node.tier !== 0 ) ) {
-      return ui.notifications.error("BACKGROUND.ERRORS.TalentTier", {localize: true});
+      return ui.notifications.error(_loc("BACKGROUND.ERRORS.TalentTier"));
     }
     const tier = talent.system.nodes.reduce((minTier, node) => (minTier < node.tier) ? minTier : node.tier, Infinity);
     const updateData = {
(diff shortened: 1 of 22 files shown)


module/applications/sheets/item-actor-details-sheet.mjs
```diff
@@ -139,11 +139,11 @@ export default class CrucibleActorDetailsItemSheet extends CrucibleBaseItemSheet
     if ( talents.some(({item}) => item === data.uuid) ) return;
     const talent = await fromUuid(data.uuid);
     if ( talent?.type !== "talent" ) {
-      ui.notifications.warn("ITEM.WARNINGS.NotTalent", {localize: true});
```
aaclayton: I don't know that this stylistic change is necessarily preferred. I suppose it does produce a shorter line of code. I'm fine with it.

aaclayton (approved):

module/applications/sheets/item-actor-details-sheet.mjs
roth-michael: I was originally going to leave it be, then saw that we already used `_loc` inside these elsewhere. Plus, had to for any which used to be `i18n.format`. But now in theory we can just ctrl+f `_loc` to get _all_ the times we localize in js (and `{{localize` for all the handlebars localization)

roth-michael:
I may do a follow up where I `ui.notify` with just the message, and `console.warn` with the stack. Feels a little "devvy" to have `Error: blah` as user facing warning banners

PR #744, opened 2026-03-15
[#743] Fix missing weapon choice for some offhand scenarios, re-add weapon context tags to action use dialog

### LINKED ISSUE #743: Having nothing (unarmed) or a shield in offhand breaks non-mainhand-tagged strikes
https://discord.com/channels/170995199584108546/1116696333894570094/1482481483796054138
`strike`, unless otherwise tagged, defaults to lowest-cost, and weapon choice is not allowed for offhand unarmed or shields.

roth-michael:
Closes #743 

Changes:
- `weapons.hasChoice` is now only not true if no natural weapons and a two-handed weapon is wielded
- If `strict` not passed to `getValidWeaponChoices`, all possible weapons are returned (besides natural - maybe we could change that), but with `isValid: false`. This lets us properly select a default weapon, for instance, in the `strike` tag's `prepare` hook
- Re-added context tags to show what weapon is being used for a `strike` (I think maybe these got accidentally lost)

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -601,6 +601,8 @@ export const TAGS = {
     prepare() {
       if ( !this.usage.weapon ) {
         const valid = this.getValidWeaponChoices({maxCost: this.actor.resources.action.value});
+
+        // Prefer mainhand weapon -> offhand weapon -> natural weapons
         for ( const weapon of valid ) {
           if ( !weapon.isValid ) continue;
           this.usage.weapon = weapon.item;
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -600,9 +600,13 @@ export const TAGS = {
     },
     prepare() {
       if ( !this.usage.weapon ) {
-        const valid = this.getValidWeaponChoices();
-        valid.sort((a, b) => a.item.system.actionCost - b.item.system.actionCost);
-        this.usage.weapon = valid[0]?.item;
+        const valid = this.getValidWeaponChoices({maxCost: this.actor.resources.action.value});
+        for ( const weapon of valid ) {
+          if ( !weapon.isValid ) continue;
+          this.usage.weapon = weapon.item;
+          break;
+        }
+        this.usage.weapon ??= valid[0]?.item;
       }
       const {strikes, weapon} = this.usage;
 
@@ -634,6 +638,8 @@ export const TAGS = {
           if ( !weaponRange ) weaponRange = weapon.system.range;
           else weaponRange = Math.min(weaponRange, weapon.system.range);
         }
+        contextTags[weapon.id] ||= {id: weapon.id, name: weapon.name, count: 0};
+        contextTags[weapon.id].count += 1;
         if ( i === 0 ) Object.assign(this.usage.bonuses, weapon.system.actionBonuses);
       }
 
@@ -649,7 +655,8 @@ export const TAGS = {
       // Context tags
       Object.assign(this.usage.context, {label: "Strikes", icon: "fa-solid fa-swords", tags: {}});
       for ( const v of Object.values(contextTags) ) {
-        this.usage.context.tags[`weapon.${v.id}`] = v.count > 1 ? `${v.name} (x${v.count})` : v.name;
+        const count = v.count * n;
+        this.usage.context.tags[`weapon.${v.id}`] = count > 1 ? `${v.name} (x${count})` : v.name;
       }
 
       // Configure action range
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -1376,7 +1376,7 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
    * @param {object} [options]              Additional options
    * @param {boolean} [options.strict]      Whether to filter out invalid items or only mark them invalid
    * @param {number|null} [options.maxCost] If provided, consider weapons with greater action cost invalid
-   * @returns {{item: CrucibleWeaponItem, label: string}[]}
+   * @returns {{item: CrucibleWeaponItem, label: string, id: string, isValid: boolean}[]}
    */
   getValidWeaponChoices({strict=false, maxCost=null}={}) {
     const choices = [];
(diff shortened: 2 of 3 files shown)


module/models/action.mjs
```diff
@@ -1374,31 +1374,43 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
+   * @returns {{item: CrucibleWeaponItem, label: string, id: string, isValid: boolean}[]}
    */
   getValidWeaponChoices({strict=false, maxCost=null}={}) {
     const choices = [];
     if ( !["strike", "reload"].some(t => this.tags.has(t)) ) return choices;
     const {mainhand: mh, offhand: oh, natural} = this.actor.equipment.weapons;
     const isValidChoice = weapon => {
       if ( this.tags.has("reload") ) return weapon.system.needsReload;
-      let isValid = this.tags.has("ranged") && weapon.config.category.ranged && !(strict && weapon.system.needsReload);
+      let isValid = this.tags.has("ranged") && weapon.config.category.ranged && !weapon.system.needsReload;
```
aaclayton, line 1387: Why do we remove the `strict` check here when determining whether weapon loaded state is valid or not?

module/models/action.mjs
```diff
@@ -1374,31 +1374,43 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
+        item: oh,
+        id: oh.id || "offhandUnarmed",
+        label: `${oh.name} (${SYSTEM.WEAPON.SLOTS.labels.OFFHAND})`,
+        isValid
+      });
     }
     if ( !this.tags.has("ranged") ) {
       for ( const n of natural ) {
-        choices.push({item: n, id: n.id, label: `${n.name} (${SYSTEM.WEAPON.PROPERTIES.natural.label})`});
+        choices.push({item: n, id: n.id, label: `${n.name} (${SYSTEM.WEAPON.PROPERTIES.natural.label})`, isValid: true});
```
aaclayton: We skip the `maxCost` check here that we make for other weapons. That seems undesirable. I would imagine that affordability would rule out natural weapons in `strict` mode.

module/models/actor-base.mjs
```diff
@@ -676,7 +676,7 @@ export default class CrucibleBaseActor extends foundry.abstract.TypeDataModel {
     weapons.dualWield = weapons.unarmed || (mh?.id && oh?.id && !weapons.shield);
     weapons.dualMelee = weapons.dualWield && !mhCategory.ranged && !ohCategory.ranged;
     weapons.dualRanged = weapons.dualWield && mhCategory.ranged && ohCategory.ranged;
-    weapons.hasChoice = weapons.dualWield || (weapons.natural.length > 0) || (weapons.melee && weapons.ranged);
+    weapons.hasChoice = (weapons.natural.length > 0) || !weapons.twoHanded;
```
aaclayton, line 679: Confirming my understanding that this will result in a choice for an actor with a 1h weapon equipped and no offhand (or an offhand only and no mainhand) where they can choose their unarmed strike?

module/const/action.mjs
```diff
@@ -600,9 +600,13 @@ export const TAGS = {
...
-        const valid = this.getValidWeaponChoices();
-        valid.sort((a, b) => a.item.system.actionCost - b.item.system.actionCost);
-        this.usage.weapon = valid[0]?.item;
+        const valid = this.getValidWeaponChoices({maxCost: this.actor.resources.action.value});
+        for ( const weapon of valid ) {
+          if ( !weapon.isValid ) continue;
+          this.usage.weapon = weapon.item;
+          break;
+        }
+        this.usage.weapon ??= valid[0]?.item;
```
aaclayton, line 611: Previously the fallback weapon was sorted as the cheapest one. Confirming that it's intentional that we choose `valid[0]` in declaration order? Might be worth a comment.

module/models/action.mjs
```diff
@@ -1374,31 +1374,43 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
   getValidWeaponChoices({strict=false, maxCost=null}={}) {
     const choices = [];
     if ( !["strike", "reload"].some(t => this.tags.has(t)) ) return choices;
     const {mainhand: mh, offhand: oh, natural} = this.actor.equipment.weapons;
     const isValidChoice = weapon => {
       if ( this.tags.has("reload") ) return weapon.system.needsReload;
-      let isValid = this.tags.has("ranged") && weapon.config.category.ranged && !(strict && weapon.system.needsReload);
+      let isValid = this.tags.has("ranged") && weapon.config.category.ranged && !weapon.system.needsReload;
       isValid ||= this.tags.has("melee") && !weapon.config.category.ranged;
       if ( maxCost !== null ) isValid &&= weapon.system.actionCost <= maxCost;
```
aaclayton:
```suggestion
      if ( maxCost !== null ) isValid &&= (weapon.system.actionCost <= maxCost);
```

aaclayton (changes requested):
Looks good! A few callouts and minor tweaks before merging

module/models/action.mjs
roth-michael: We changed `strict` to mean "should invalid items be filtered out, or simply returned with `isValid: false`, so it no longer made sense to have that inside this `isValidChoice` function.

module/models/actor-base.mjs
roth-michael: Correct. Will also allow someone with mainhand and offhand _shield_ to smack with shield, if desired.

module/const/action.mjs
roth-michael: Confirmed intentional - until/unless we implement some `lastStrike` which tracks what the last weapon used for a strike was, we should always prefer mainhand over offhand, as the alternative would be (for a one-handed, nothing-in-offhand actor) an always-cheaper cost of the offhand unarmed strike, which isn't great UX. I'll add a comment.

aaclayton (approved):

PR #747, opened 2026-03-16
[#746] Update actor details items if necessary during `syncTalents`

### LINKED ISSUE #746: Ensure talent ID migration is accounted for with existing actors' detail items
When we modify talent IDs which are granted by existing Ancestries, Archetypes, Backgrounds or Taxonomies, we also adjust those detail items items themselves in the compendia to use the new uuids in their `system.talents` arrays. This, however, does nothing for already-existing actors. Their talents themselves will be migrated, but their synthetic detail items (e.g. `system.details.background`) will not be updated with new talents based on its component parts.

We should account for this in `syncOwnedItems`/`syncTalents`.

### LINKED ISSUE #769: Optimize repeated `CrucibleActor#<operation>EmbeddedDocuments` calls using new `modifyBatch`
There are a number of spots where we collect a list of documents to create, update, and delete, then call the respective functions to do so on an actor (sometimes also updating the actor itself). With V14, we can batch these into a single call with `foundry.documents.modifyBatch`.

Added `CrucibleActor#defineBatchOperations`:
```js
/**
 * @typedef {"createItems"|"updateItems"|"deleteItems"|"createEffects"|"updateEffects"
 * |"deleteEffects"} BatchOperationKey
 */
/**
 * @typedef {Record<BatchOperationKey, Array|{changes: Array, options: object}} BatchOperationEmbeddedChanges
 */
/**
 * Create a batch operation list to be used with modifyBatch
 * @param {object} actorUpdates                           Direct actor updates to include in the batch operation
 * @param {BatchOperationEmbeddedChanges} embeddedChanges Embedded document changes to include in the batch operation
 * @returns {DatabaseWriteOperation[]}
 */
defineBatchOperations(actorUpdates, embeddedChanges) {
```

We now make use of this in:
- `CrucibleActor##applyOutcomeEffects`
- `CrucibleActor##applyActiveEffectChanges` (as a note, this is potentially a confusingly-named function)
- `CrucibleActor#syncTalents`
- `CrucibleActor#syncIconicSpells`
- `CrucibleActor##grantDetailTalents`
- `crucible.api.methods.syncOwnedItems`

roth-michael:
Closes #746

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1640,11 +1640,13 @@ export default class CrucibleActor extends Actor {
       const talents = [];
       let needsUpdate = false;
       for ( const {item, level} of oldTalents ) {
-        const talentId = item.slice(-16);
-        const migratedId = migrations[talentId]?.slice(-16);
-        if ( migratedId ) {
+        const talentId = foundry.utils.parseUuid(item).id;
+        const migratedUuid = migrations[talentId];
+
+        // If undefined, no migration. If null, talent was deleted, and we should remove. Otherwise use new uuid
+        if ( migratedUuid !== undefined ) {
           needsUpdate ||= true;
-          talents.push({item: item.replace(talentId, migratedId), level});
+          if ( migratedUuid ) talents.push({item: migratedUuid, level});
           continue;
         }
         talents.push({item, level});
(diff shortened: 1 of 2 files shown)


module/documents/actor.mjs
```diff
@@ -1631,14 +1632,35 @@ export default class CrucibleActor extends Actor {
...
 
+    // Ensure all details "items" are using migrated versions of uuids
+    const detailItemTypes = ["ancestry", "archetype", "background", "taxonomy"];
+    for ( const detailType of detailItemTypes ) {
+      const oldTalents = this.system.details[detailType]?.talents;
+      if ( !oldTalents ) continue;
+      const talents = [];
+      let needsUpdate = false;
+      for ( const {item, level} of oldTalents ) {
+        const talentId = item.slice(-16);
```
aaclayton: Use `foundry.utils.parseUuid`

module/documents/actor.mjs
```diff
@@ -1631,14 +1632,35 @@ export default class CrucibleActor extends Actor {
...
+        const migratedId = migrations[talentId]?.slice(-16);
```
aaclayton: Careful here. `TALENT_ID_MIGRATIONS` uses `null` to indicate "this talent was deleted", but you are keeping such talents in the `talents` array rather than removing them.

module/documents/actor.mjs
```diff
@@ -1631,14 +1632,35 @@ export default class CrucibleActor extends Actor {
...
+        if ( migratedId ) {
+          needsUpdate ||= true;
```
aaclayton, line 1654: Redundant

module/documents/actor.mjs
```diff
@@ -1631,14 +1632,35 @@ export default class CrucibleActor extends Actor {
...
+          talents.push({item: item.replace(talentId, migratedId), level});
```
aaclayton: Aren't the values of `TALENT_ID_MIGRATIONS` already full UUIDs that would be used directly? (unless `null`)

aaclayton (changes requested):
Thanks! a few comments

module/documents/actor.mjs
roth-michael: Do you mean that tracking `needsUpdate` at all is unnecessary?

aaclayton: Aah, I think I misread that this was used in the *outer loop* to test whether any talent had a `migratedId`. I withdraw my comment.

aaclayton (approved):
The batching changes look nice, lots of places to use this approach.

PR #749, opened 2026-03-17
Remove unclear references to Finesse/Brute, fix some tag rules & i18n

roth-michael:
Changes to character mechanics, because github has those big sections instead of the precise changes:
Before:
> Strength affects Fortitude defense, carrying Capacity, and the use of weaponry __with the Brute trait__.
> Dexterity affects Initiative in taking action, the Dodge component of Physical Defense, Reflex defense, and the use of weaponry __with the Finesse trait__.

After:
> Strength affects Fortitude defense, carrying Capacity, and the use of weaponry __which scales with Strength__.
> Dexterity affects Initiative in taking action, the Dodge component of Physical Defense, Reflex defense, and the use of weaponry __which scales with Dexterity__.

### DIFF
--- a/_source/talent/Weak_Points_weakpoints000000.yml
+++ b/_source/talent/Weak_Points_weakpoints000000.yml
@@ -5,10 +5,9 @@ img: icons/skills/melee/strike-sword-stabbed-brown.webp
 system:
   description: >-
     <p>You identify moments of opportunity to strike at an enemy's weak points.
-    When attacking a target that is <strong>Unaware</strong>,
-    <strong>Flanked</strong>, or <strong>Exposed</strong> using a weapon which
-    scales with <strong>Dexterity</strong>, you gain +2 damage to any
-    Strike.</p>
+    When attacking a target that is @Condition[unaware], @Condition[flanked], or
+    @Condition[exposed] using a weapon which scales with
+    <strong>Dexterity</strong>, you gain +2 damage to any Strike.</p>
   actions: []
   nodes:
     - dex0b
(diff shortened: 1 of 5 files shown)


lang/en.json
```diff
@@ -250,13 +250,13 @@
       "Deadly": "Deadly",
       "DeadlyTooltip": "This action has the potential to deal a tremendous amount of damage, increasing the damage multiplier for any damage dealt by x1.",
```
aaclayton, line 251: This changed in the rules text but not here, should be made consistent.

_source/talent/Weak_Points_weakpoints000000.yml
```diff
@@ -4,10 +4,11 @@ _id: weakpoints000000
 img: icons/skills/melee/strike-sword-stabbed-brown.webp
 system:
   description: >-
-    You identify moments of opportunity to strike at an enemy's weak points.
+    <p>You identify moments of opportunity to strike at an enemy's weak points.
     When attacking a target that is <strong>Unaware</strong>,
-    <strong>Flanked</strong>, or <strong>Exposed</strong> using Finesse
-    weaponry, you gain +2 damage to any Strike.
+    <strong>Flanked</strong>, or <strong>Exposed</strong> using a weapon which
```
aaclayton: Use `@Condition[id]` for unaware, flanked, and exposed instead of `<strong>` tags.

aaclayton (approved):
Approved, but minor edits to please make before I merge it.

lang/en.json
roth-michael: I'm not sure whether it's clearer in the rules text or in this tooltip text. If you think rules text is better then I'll make the change here. Otherwise, I think a separate PR ought to be tidying up the whole "action tags" section of the rules journal.

aaclayton (approved):

PR #753, opened 2026-03-19
[WIP] minimal implementation of stack-aware toolbelt

### LINKED ISSUE #685: Prevent stacking of consumables on toolbelt
Or perhaps set a "max in toolbelt slot" value on consumables - maybe a question of intended balance there.

danielrab:
my proposed solution for https://github.com/foundryvtt/crucible/issues/685


roth-michael:
A UX concern here - if you have a stack of items which is greater than your open toolbelt slots, there's no easy way to "split" one out for your toolbelt.

danielrab:
Yeah, I guess we do need a way to split and combine stacks, and preferably when you drag a stack that is too big, it should automatically split exactly enough to fill the available slots.
Also, a more minor UX concern is that you can go over the limit by changing the size of an existing stack, as that's currently an unchecked action.

module/applications/sheets/base-actor-sheet.mjs
```diff
@@ -449,7 +449,8 @@ export default class CrucibleBaseActorSheet extends api.HandlebarsApplicationMix
 
     // Sort inventory
     for ( const heading of Object.values(sections.inventory) ) {
-      if ( heading.counter ) heading.label += ` (${heading.items.length}/${heading.counter})`;
+      const totalQuantity = heading.items.reduce((acc, item) => acc + item.quantity, 0);
+      if ( heading.counter ) heading.label += ` (${totalQuantity}/${heading.counter})`;
```
roth-michael:
```suggestion
      if ( heading.counter ) {
        const totalQuantity = heading.items.reduce((acc, item) => acc + item.quantity, 0);
        heading.label += ` (${totalQuantity}/${heading.counter})`;
      }
```
No need to count up quantities if we're not putting them in the header.

module/documents/actor.mjs
```diff
@@ -2113,9 +2113,7 @@ export default class CrucibleActor extends Actor {
     slot = (equipped && !stowDropped) ? result.slot : undefined;
     const action = equipped ? this.#equipItemAction(item, slot) : this.#unequipItemAction(item, dropped);
     if ( this.inCombat ) await action.use();
-    else if ( action.usage.actorUpdates.items.length ) {
-      await this.updateEmbeddedDocuments("Item", action.usage.actorUpdates.items);
-    }
+    else await this.update(action.usage.actorUpdates);
```
roth-michael, line 2164: This makes sense. Commenting for clarity: if we're creating a new item in `action.usage.actorUpdates.items` then we can't use `updateEmbeddedDocuments`. 

I do wonder whether it wouldn't be better to be explicit about item creation/update/deletion, but I don't think there's anything _wrong_ with this approach, if we are indeed going to be storing the item data in `usage`.

module/documents/actor.mjs
```diff
@@ -2173,7 +2171,24 @@ export default class CrucibleActor extends Actor {
...
+      console.log(mergableItem);
+      if (mergableItem) {
+        const mergeUpdate = {_id: mergableItem._id, system: {quantity: mergableItem.system.quantity + 1}};
+        action.usage.actorUpdates.items.push(mergeUpdate);
+      } else {
+        action.usage.actorUpdates.items.push(splitOff);
+      }
+    } else {
+      update.system = {dropped: false, equipped: true};
+    }
```
roth-michael, line 2239: I've made some local changes here to fix up some edge cases & see if I can make it a bit more robust. Will further explain in the full review comment. I'd be hesitant to move forward until we can double check with Andrew whether this mechanism of item creation is alright, or if he'd prefer that we do new handling where we instead only store the item UUID and perform this logic at the time of confirmation.

roth-michael (commented):
Good general concept, I think. As I mentioned in the third review comment, I think we'll want to evaluate storing item data in the action vs storing the uuid and performing the logic for merging at the time of confirmation. The changes I made locally:
In `CrucibleActor##equipItemAction`:
```js
// Equip the weapon as a follow-up actor update
action.usage.actorUpdates ||= {};
action.usage.actorUpdates.items ||= [];
const update = {_id: item.id};
if ( item.system.properties.has("stackable") ) { // <-- change
  this.#handleItemSplit(item, action, update, true); // <-- change
} else {
  update.system = {dropped: false, equipped: true};
}
```
In `CrucibleActor##unequipItemAction` (pretty much all a change):
```js
action.usage.actorUpdates ||= {};
action.usage.actorUpdates.items ||= [];
const update = {_id: item.id};
if ( item.system.properties.has("stackable") ) {
  this.#handleItemSplit(item, action, update, false);
} else {
  update.system = {dropped: !!dropped, equipped: false};
  if ( item.type === "weapon" ) Object.assign(action.usage.actorStatus, {unequippedWeapon: true});
}
action.usage.actorUpdates.items.push(update);
return action;
```
A new `CrucibleActor##handleItemSplit` (which would need fixing up, I think):
```js
#handleItemSplit(item, action, update, equipped) {
  const splitOff = item.clone({system: {quantity: 1, dropped: false, equipped}});
  const mergeableItem = this.items.find(candidate => candidate.isStackableWith(splitOff));
  if ( mergeableItem ) {
    update.system = {quantity: item.system.quantity - 1};
    const mergeUpdate = {_id: mergeableItem._id, system: {quantity: mergeableItem.system.quantity + 1}};
    action.usage.actorUpdates.items.push(mergeUpdate);
  } else if ( item.system.quantity > 1 ) {
    update.system = {quantity: item.system.quantity - 1};
    action.usage.actorUpdates.items.push(splitOff.toObject());
  } else {
    update.system = {dropped: false, equipped};
  }
}
```
and, as referenced in that new function, instead of `CrucibleItem#toCleanData`:
```js
isStackableWith(other) {
  const cleanData = this.toObject();
  const cleanOther = other.toObject();
  for ( const data of [cleanData, cleanOther] ) {
    delete data._id;
    delete data._stats;
    delete data.system.quantity;
    delete data.sort;
    data.system.dropped = !!data.system.dropped;
  }
  return foundry.utils.equals(cleanData, cleanOther);
}
```

All that, combined with an early return in `CrucibleBaseActorSheet##preparePhysicalItem` in the case that `system.quantity === 0` (which stops rendering of depleted items) should be solid. Then just need to handle cleanup of said hidden depleted items. 

I don't know whether it makes sense to actually implement any of those changes on this PR currently - I think we should consider this blocked until Atro has time to consider how he'd like this implemented.

module/documents/actor.mjs
danielrab: It also mimics more closely what the confirmation of an action does, making it less likely for the two implementations to desync in any way.

danielrab:
to be clear, I know the current implementation doesn't handle all cases yet. Even if there are no problems with what's there, it's not entirely ready to merge yet.

aaclayton:
Closed in favor of https://github.com/foundryvtt/crucible/commit/27cdd3e7d3352d9b8c7ab422ef733b2f3449785f

PR #754, opened 2026-03-19
[#725] Update token ruler logic for clearer info presentation, fix non-direct movement bug

### LINKED ISSUE #725: Improve the presentation of the Crucible Token Ruler waypoints, fixing several bugs with incorrect labeling along the way
For the sake of simplicity, assume for these examples that any movement action is confirmed prior to the next movement action taking place, the actor in question has a 10ft stride, is in combat, and hasn't free-moved yet.

**Issue #1**
1. Move actor 5ft. Confirm. 0 Action. Hovering the actor shows the waypoint label: `5ft 0A = [0A]`
2. Start dragging the actor, but cancel (right click or release at same spot). Waypoint label now says: `5ft 1A = [1A]`

**Issue #2**
1. Move actor 5ft. Confirm. 0 Action. Hovering the actor shows the waypoint label: `5ft 0A = [0A]`
2. Drag the actor out to 10ft total, don't yet release. Waypoint preview shows: `10ft (+5) 0A = [1A]` - it should read `10ft (+5) 1A = [1A]`
3. Completing that movement & confirming _does_ properly cost 1A, but hovering _now_ shows the first waypoint: `5ft 1A` and the second `10ft (+5) 0A = [1A]`
4. Similarly, moving 5ft and then 5ft again (separate movement actions) shows `15ft (+5) 1A`, then `20ft (+5) 0A = [2A]` though, again, the actual actions consumed is correct (1 each)

Tried looking into this a bit, though I really still haven't had to wrap my head around the specifics of movement in v13+.

I believe that the first issue happens because `CrucibleActor#getMovementActionCost` is called when getting the waypoint label, and when called after the free movement has been "used," it'll return what the cost _would have_ been without free movement.

The second issue, I think, compounds on the first issue. The previous movement, it thinks, cost 1A. Therefore any additional movement within stride, it determines, shouldn't cost any extra. (Again, this is all visual, the actual calculations are done properly).

roth-michael:
Closes #725 
Fixes an issue mentioned in the bug reports channel where normal movement could not be waypointed.

Will not be fully functional until V14 Testing 2 (requires https://github.com/foundryvtt/foundryvtt/issues/13985)

Example of multiple varying-in-complexity movements:
image

### DIFF
--- a/module/canvas/token-ruler.mjs
+++ b/module/canvas/token-ruler.mjs
@@ -14,19 +14,54 @@ export default class CrucibleTokenRuler extends foundry.canvas.placeables.tokens
     const grid = canvas.scene?.grid;
     if ( !grid || (grid.distance !== 1) || (grid.units !== "ft") ) return;
 
-    // Recover prior cost (potentially across chains)
-    state.priorCost ??= actor.getMovementActionCost(waypoint.previous?.measurement.cost || 0).cost;
+    // Get measurement of prior subpath, if on new subpath
+    state.priorCost ??= 0;
+    state.priorDistance ??= 0;
+    if ( state.priorSubpathId !== waypoint.subpathId ) {
+      let previous = waypoint.previous;
+      while ( previous ) {
+        if ( previous.subpathId !== waypoint.subpathId ) break;
+        previous = previous.previous;
+      }
+      state.priorCost = waypoint.previous?.measurement.cost || 0;
+      state.priorDistance = waypoint.previous?.measurement.distance || 0;
+      state.priorSubpathId = waypoint.subpathId;
+    }
 
-    // Measure new segment
-    const movement = actor.getMovementActionCost(waypoint.measurement.cost);
-    const deltaCost = movement.cost - state.priorCost;
-    state.priorCost = movement.cost;
+    // Do not draw waypoint label for non-meaningful waypoints. Meaningful waypoints:
+    // - Are the final waypoint of a subpath
+    // - Change movement action
+    // - Change elevation
+    const isFinalOfSubpath = waypoint.next?.subpathId !== waypoint.subpathId;
+    const sameAction = waypoint.previous?.action === waypoint.action;
+    const sameElevation = waypoint.previous?.elevation === waypoint.elevation;
+    if ( !isFinalOfSubpath && sameElevation && sameAction) return;
 
-    context.distance.units = grid.units;
-    context.cost = {units: "A", delta: deltaCost};
-    context.cost.total = Number.isFinite(movement.cost) ? movement.cost : "Impossible";
-    context.cost.showTotal = !waypoint.next;
-    context.displayElevation = context.elevation && !context.elevation.hidden;
+    // Calculate cost of this subpath & set context appropriately
+    const movement = actor.getMovementActionCost(waypoint.measurement.cost - state.priorCost);
+    const freeMovementId = actor.system.status.freeMovementId;
+    const deltaCost = movement.cost - ((freeMovementId && (freeMovementId === waypoint.subpathId)) ? 1 : 0);
+    Object.assign(context.distance, {
+      units: grid.units,
+      total: (context.distance.total - state.priorDistance).toNearest(0.01).toLocaleString(game.i18n.lang)
+    });
+    context.cost = {
+      units: "A",
+      delta: Number.isFinite(deltaCost) ? deltaCost : "Impossible",
+      displayCost: isFinalOfSubpath
+    };
+    context.displayElevation = context.elevation && !context.elevation.hidden && (!sameElevation || waypoint.elevation);
     return context;
+  }
+
+  /** @override */
+  _getWaypointStyle(waypoint) {
+    const style = super._getWaypointStyle(waypoint);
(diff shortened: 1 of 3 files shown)


module/canvas/token-ruler.mjs
```diff
@@ -14,19 +14,54 @@ export default class CrucibleTokenRuler extends foundry.canvas.placeables.tokens
...
-    context.displayElevation = context.elevation && !context.elevation.hidden;
+    // Get measurement of prior subpath, if on new subpath
+    state.priorCost ??= 0;
+    state.priorDistance ??= 0;
+    if ( state.priorSubpathId !== waypoint.subpathId ) {
+      let previous = waypoint.previous;
+      while ( previous ) {
+        if ( previous.subpathId !== waypoint.subpathId ) break;
+        previous = previous.previous;
+      }
```
aaclayton: Perhaps this was an artifact of prior iteration, or perhaps it represents a bug, but this while loop to identify `previous` doesn't end up being used and is dead code. In the lines that immediately follow you are reading from `waypoint.previous` directly.

Assuming the current code is *working* properly, then I think these lines can just be deleted.

module/canvas/token-ruler.mjs
```diff
@@ -14,19 +14,54 @@ export default class CrucibleTokenRuler extends foundry.canvas.placeables.tokens
...
+    }
+
+    // Do not draw waypoint label for non-meaningful waypoints. Meaningful waypoints:
+    // - Are the final waypoint of a subpath
+    // - Change movement action
+    // - Change elevation
+    const isFinalOfSubpath = waypoint.next?.subpathId !== waypoint.subpathId;
+    const sameAction = waypoint.previous?.action === waypoint.action;
+    const sameElevation = waypoint.previous?.elevation === waypoint.elevation;
+    if ( !isFinalOfSubpath && sameElevation && sameAction) return;
```
aaclayton:
```suggestion
    if ( !isFinalOfSubpath && sameElevation && sameAction ) return;
```

module/canvas/token-ruler.mjs
```diff
@@ -14,19 +14,54 @@ export default class CrucibleTokenRuler extends foundry.canvas.placeables.tokens
...
+
+    // Calculate cost of this subpath & set context appropriately
+    const movement = actor.getMovementActionCost(waypoint.measurement.cost - state.priorCost);
+    const freeMovementId = actor.system.status.freeMovementId;
+    const deltaCost = movement.cost - ((freeMovementId && (freeMovementId === waypoint.subpathId)) ? 1 : 0);
```
aaclayton, line 40: Want to be careful about something here.

`actor.getMovementActionCost` is called without `{useFreeMove: false}`, so it's going to apply a free move discount internally if `actor.system.hasFreeMove`.

We then are subtracting 1 from `deltaCost` if there is a `freeMovementId` that matches this preexisting `subpathId`. 

Maybe these are paths that never cross, but I'm just a little worried looking at this code whether we are somehow allowing a double free move? If it's just me misreading the logic, maybe different variables names or comment lines could clarify how this is working.

module/canvas/token-ruler.mjs
```diff
@@ -14,19 +14,54 @@ export default class CrucibleTokenRuler extends foundry.canvas.placeables.tokens
...
+    Object.assign(context.distance, {
+      units: grid.units,
+      total: (context.distance.total - state.priorDistance).toNearest(0.01).toLocaleString(game.i18n.lang)
+    });
+    context.cost = {
+      units: "A",
+      delta: Number.isFinite(deltaCost) ? deltaCost : "Impossible",
+      displayCost: isFinalOfSubpath
+    };
+    context.displayElevation = context.elevation && !context.elevation.hidden && (!sameElevation || waypoint.elevation);
```
aaclayton: Just some commentary for posterity:

```suggestion

    // Display elevation when it changes or when different from the level.elevation.bottom
    context.displayElevation = context.elevation && !context.elevation.hidden && (!sameElevation || waypoint.elevation);
```

We need to be careful here about the units of `waypoint.elevation`. I think maybe its absolute elevation for the Scene, but I think our labels should present elevation relative to the current `level.elevation.bottom` the same way that the Token HUD does. 

Maybe test this on the "Multi-Level Keep" module that we've been providing for V14 testing of scene levels and verify the behavior when you are on a level that has `elevation.level.bottom !== 0`.

aaclayton (changes requested):
Looks good, just a couple comments and questions to get comfortable with

module/canvas/token-ruler.mjs
roth-michael: Huh. Yeah, vestigial I think, from when we wanted to show total cost (including prior movements) in addition to "delta." Will remove.

module/canvas/token-ruler.mjs
roth-michael: Paths won't cross because if `actor.system.status.freeMovementId` actually exists, that means that the free movement has been confirmed, and so `actor.system.hasFreeMove` will be false. But I'll throw in a comment to clarify.

aaclayton (approved):

PR #755, opened 2026-03-21
Update Crucible movement planning to use new upstream v14 API

aaclayton:
V14 added a high-level `Token#planMovement` API, which can replace much of the custom machinery I had to build to do this before. 

@roth-michael could you help review this code and do some local testing to verify that movement (both unplanned and planned) is working as expected?

This might also intersect with #754 in a way that makes sense to do one before the other (unsure which order would be better).

### DIFF
--- a/module/dice/action-use-dialog.mjs
+++ b/module/dice/action-use-dialog.mjs
@@ -32,25 +32,10 @@ export default class ActionUseDialog extends StandardCheckDialog {
   static TEMPLATE = "systems/crucible/templates/dice/action-use-dialog.hbs";
 
   /**
-   * A registry of ActionUseDialog instances that are currently in a movement planning state.
-   * @type {Map<string, ActionUseDialog>}
-   */
-  static #movementPlans = new Map();
-
-  /* -------------------------------------------- */
-
-  /**
    * Targets acquired from the most recently placed region for this Action.
    * @type {ActionUseTarget[]|null}
    */
   #regionTargets = null;
-
-  /**
-   * The Hooks ID for a pending planToken listener registered during movement planning.
-   * Stored so it can be cleaned up if the dialog closes before planning completes.
-   * @type {number|null}
-   */
-  #planMovementHookId = null;
 
   /**
    * A lazy action clone used for per-frame target preview during movement planning.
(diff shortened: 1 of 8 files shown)


module/dice/action-use-dialog.mjs
```diff
@@ -529,36 +507,38 @@ export default class ActionUseDialog extends StandardCheckDialog {
...
-        this.#planMovementHookId = null;
-        resolve();
-      });
+    // Await the movement plan
+    const plan = await token.object.planMovement({
+      allowedActions: movementUsage.action ? [movementUsage.action] : null,
+      minCost: this.action.range?.minimum ?? undefined,
+      maxCost: this.action.range?.maximum ?? undefined,
+      direct: movementUsage.direct ?? true,
+      constrainOptions: movementUsage.ignoreWalls ? {ignoreWalls: true} : undefined
```
roth-michael: Is there any chance we'd want to add other `constrainOptions` in the future? I guess in that case we could just add `ignoreCost` (for example) to `movement.usage` and construct the actual `constrainOptions` prior to this call.

module/documents/actor.mjs
```diff
@@ -408,14 +409,29 @@ export default class CrucibleActor extends Actor {
...
+      derivedActions.add(w.action);
+    }
+    for ( const w of movement.pending.waypoints ) {
+      waypoints.push(w);
+      derivedActions.add(w.action);
+    }
+    actions ??= derivedActions;
 
     // Annotate movement actions.
     actions ||= new Set(["walk"]);
```
roth-michael: This'll never do anything, `||=` considers an empty set to be truthy. Depending on whether we want empty set to mean "use walk" or to mean "derive actions" we'd either want this line to be 
```diff
- actions ||= new Set(["walk"]);
+ if ( !actions.size ) actions.add("walk");
```
or change 434 to be
```diff
- actions ??= derivedActions;
+ if ( !actions?.size ) actions = derivedActions;
```
and remove this line, respectively

module/models/action.mjs
```diff
@@ -2069,18 +2067,30 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
+      else {  // Reconstruct movement data from history
+        let firstIdx;
+        waypoints = token.movementHistory.filter((w, i) => {
+          if ( w.movementId === movementId ) {
+            firstIdx ??= i;
+            return true;
+          }
+          return false;
+        });
+        origin = token.movementHistory[firstIdx - 1];
```
roth-michael: May be me misunderstanding here, but why wouldn't the origin just be the first found match of the movement ID? What happens here if `firstIdx` is `0`? Also, with multi-waypointed movement this is will ignore all but the first waypoint. I think we wanna use `subpathId` here:

```suggestion
        waypoints = token.movementHistory.filter(w => w.subpathId === movementId);
        origin = waypoints[0];
```

roth-michael (changes requested):
Looks good overall to me, testing yields no unexpected behavior. 2/3 review comments I think should be addressed, one is just thinking out loud.

Some stuff we should still cover, I think:
- After planning movement, nothing prevents moving the token (with the action use dialog still open)
- After posting the chat message for a planned-movement action, nothing prevents moving the token, nor does moving the token remove the ability to confirm the planned action (though confirming doesn't result in moving the token)

I think probably we can get this in _before_ #754 and I can adjust it as needed (hopefully nothing will be needed).

PR #763, opened 2026-03-26
Effects standardization

### LINKED ISSUE #764: standardize customizability of provided effects functions
Most, but not all effect functions in `effect.mjs` accept optional parameters to change their duration, scaling or strength.
This should be standardized so that people know what to expect when making homebrew items.

danielrab:
make all effect functions optionally accept turns, amount, and ability (whenever applicable)
stop passing a target argument that wasn't used
consistently use actor.getAbilityBonus

### DIFF
--- a/module/const/effects.mjs
+++ b/module/const/effects.mjs
@@ -252,7 +252,7 @@ export function mending(actor, {ability="wisdom", amount, turns=1}={}) {
  * @param {Actor} actor
  * @param {Actor} target
  */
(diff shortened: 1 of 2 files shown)


roth-michael (approved):
LGTM. As discussed in discord, only potential question for _now_ is whether to allow `damageType` for all DoT effects (I think no). And a question for later on whether to change `turns` to `duration` or similar.

aaclayton (approved):

PR #768, opened 2026-03-28
[#767] Rework `crucible.methods.generateId` and `SYSTEM.EFFECTS.getEffectId`, fix misc bugs

### LINKED ISSUE #767: Miscellaneous 0.9.1 bugfixes
- Hamstring throws an error if using a two-handed weapon
- Vampiric Bite does not account for natural weapons training
- Armored Shell throws an error if using a two-handed weapon
- Adrenaline Surge does not recognize the effect
- Vow of Animus does not recognize the effect

roth-michael:
Closes #767 
For Hamstring & Armored Shell, optional chain for offhand. For Vampiric Bite, call `prepareEquippedData` on the data model of the synthetic weapon. For Adrenaline Surge & Vow of Animus, rework `generateId` and `getEffectId`:
- `generateId` now preserves casing
- `getEffectId` now calls `generateId`
- `getEffectId` now accepts an optional `suffix` which replaces the last `n` characters of the returned string with `suffix` (where `n` is the length of `suffix`)
- `getEffectId` with the `suffix` argument is now used when assigning effect `_id`s in `CrucibleAction##attachOutcomeEffects`

I would be down to rename `generateId` to `slugify`, and potentially move `getEffectId` into `crucible.api.methods`, and rename it to `generateId`.

### DIFF
--- a/module/const/effects.mjs
+++ b/module/const/effects.mjs
@@ -9,11 +9,12 @@
 
 /**
  * Get a standardized 16 character ID that can be used for the ActiveEffect.
- * @param {string} label    The active effect label
- * @param {string} [suffix] A suffix with which to end the returned ID
- * @returns {string}        The standardized ID
+ * @param {string} label            The active effect label
+ * @param {object} [options]        Additional options
+ * @param {string} [options.suffix] A suffix with which to end the returned ID
+ * @returns {string}                The standardized ID
  */
-export function getEffectId(label, suffix="") {
+export function getEffectId(label, {suffix=""}={}) {
   return `${crucible.api.methods.generateId(label, 16 - suffix.length)}${suffix}`;
 }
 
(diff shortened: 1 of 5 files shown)


module/const/effects.mjs
```diff
@@ -10,10 +10,11 @@
 /**
  * Get a standardized 16 character ID that can be used for the ActiveEffect.
  * @param {string} label    The active effect label
+ * @param {string} [suffix] A suffix with which to end the returned ID
  * @returns {string}        The standardized ID
  */
-export function getEffectId(label) {
-  return label.slugify({replacement: "", lowercase: false, strict: true}).slice(0, 16).padEnd(16, "0");
+export function getEffectId(label, suffix="") {
```
aaclayton: I'd prefer we use an `options` object, never know when we might need another option and it sucks to have to extend or break the method signature every time.

aaclayton (changes requested):

aaclayton (approved):

PR #775, opened 2026-03-30
[773] Update text enricher group checks to new API with prefilled actors

### LINKED ISSUE #773: Update text enricher group checks to new API with prefilled actors
Refactor the TextEnricher group checks to use the new group check API and prefill it with existing party actors.

Clicking enriched content should open a roll request dialog prefilled with actors and DC, which can be triggered by the GM.

Add bones and banes as additional side targets.

Plushtoast:
Fixes #773

group enricher now prepares a group check request.

Also fixes a bug in group check boon/banes prep.

btw styling for that request card looks a little weird still.

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -618,7 +618,7 @@ function enrichSpell([match, spellId]) {
  */
 function enrichSkillCheck([match, terms]) {
   let [skillId, dc, ...rest] = terms.split(" ");
-  if ( skillId in DND5E_SKILL_MAPPING ) skillId = DND5E_SKILL_MAPPING[skillId];
+  skillId = DND5E_SKILL_MAPPING[skillId] ?? skillId;
   const skill = SYSTEM.SKILLS[skillId];
   if ( !skill ) return new Text(match);
   const passive = rest.includes("passive");
@@ -639,22 +639,25 @@ function enrichSkillCheck([match, terms]) {
 function createSkillCheckElement(skill, dc, {passive=false, group=false}={}) {
   const tag = document.createElement("enriched-content");
   tag.classList.add("skill-check", skill.category);
-  if ( group ) tag.classList.add("group-check");
-  tag.dataset.skillId = skill.id;
-  tag.dataset.dc = dc;
+  const dataset = {skillId: skill.id, dc};
+  if ( group ) {
+    tag.classList.add("group-check");
+    dataset.group = "true";
+  }
   let dcLabel = _loc("DICE.DCSpecific", {dc});
 
   // Passive checks only
   if ( passive ) {
     dcLabel = _loc("DICE.DCAdditionalPassive", {dcLabel});
     tag.classList.add("passive-check");
-    tag.dataset.crucibleTooltip = "passiveCheck";
+    dataset.crucibleTooltip = "passiveCheck";
   }
 
   // Group checks only
   if ( group ) dcLabel = _loc("DICE.DCAdditionalGroup", {dcLabel});
 
   // Create label
+  Object.assign(tag.dataset, dataset);
   tag.innerHTML = `${skill.label} (${dcLabel})`;
   return tag;
 }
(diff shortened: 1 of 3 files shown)


module/enrichers.mjs
```diff
@@ -618,12 +618,22 @@ function enrichSpell([match, spellId]) {
...
+  for ( const token of rest ) {
+    if ( token === "passive" ) options.passive = true;
+    else if ( token === "group" ) options.group = true;
+    else {
+      const [key, value] = token.split("=");
+      if ( !value ) continue;
+
+      if ( key === "boons" ) options.boons = value;
+      else if ( key === "banes" ) options.banes = value;
+    }
```
aaclayton: I'm not sure I agree about pre-filling boons or banes inline in a written skill check. I think that should be split out and scoped as a github issue for discussion before we implement. The point here was to link group checks to the group rolling UX, which is a feature that should stand on its own.

module/enrichers.mjs
```diff
@@ -618,12 +618,22 @@ function enrichSpell([match, spellId]) {
...
+  const options = {passive: false, group: false, boons: 0, banes: 0};
+  for ( const token of rest ) {
+    if ( token === "passive" ) options.passive = true;
+    else if ( token === "group" ) options.group = true;
+    else {
+      const [key, value] = token.split("=");
+      if ( !value ) continue;
+
+      if ( key === "boons" ) options.boons = value;
+      else if ( key === "banes" ) options.banes = value;
```
aaclayton: These are strings not numbers

module/enrichers.mjs
```diff
@@ -733,11 +754,20 @@ function renderSkillCheck(element) {
  */
 function onClickSkillCheck(event) {
   event.preventDefault();
-  const element = event.currentTarget;
-  const {skillId, dc} = element.dataset;
+  const {skillId, dc: dcStr, group, boons: boonsStr, banes: banesStr} = event.currentTarget.dataset;
+  const isGroupCheck = group === "true";
+  if ( isGroupCheck && !game.user.isGM ) return ui.notifications.warn(_loc("DICE.GROUP_CHECK.RequiresGM"));
+
+  const dc = Number(dcStr) || undefined;
```
aaclayton: I won't really encourage it, but I could see people setting a DC of zero for skills that you automatically succeed at if you try them. We shouldn't cause `dcStr` of `"0"` to fall back to `undefined`.

module/enrichers.mjs
```diff
@@ -733,11 +754,20 @@ function renderSkillCheck(element) {
  */
 function onClickSkillCheck(event) {
   event.preventDefault();
-  const element = event.currentTarget;
-  const {skillId, dc} = element.dataset;
+  const {skillId, dc: dcStr, group, boons: boonsStr, banes: banesStr} = event.currentTarget.dataset;
+  const isGroupCheck = group === "true";
+  if ( isGroupCheck && !game.user.isGM ) return ui.notifications.warn(_loc("DICE.GROUP_CHECK.RequiresGM"));
```
aaclayton: Could be worth factoring out something like `onClickGroupCheck()` that we delegate out to in order to keep the logic here more readable.

aaclayton (changes requested):
Looks pretty good, a few requests before we merge it.

module/enrichers.mjs
Plushtoast: Issue created:
https://github.com/foundryvtt/crucible/issues/816

Plushtoast:
ok I:

- removed the boons/banes parsing addition
- extracted the group check into its own click handler. during binding of the handler it will decide on the dataset which handler to take
- added more helper functions since single and gorup check share a lot of code
- defaulted dc to 15 for invalid dc values

the getPartyActors function is almost the same like it is on standardcheck
those two functions could also be refactored into one function somewhere else

aaclayton (approved):

PR #776, opened 2026-03-31
Restyle intermediate waypoints to be less small, but to be a different shape

roth-michael:
Core also un-translucent-ed the starting waypoint, so was able to remove that logic.

Marking as blocked because it requires at least 14.359.

Also removed the to-be-unnecessary release build check for implementing relative elevation display on the ruler.


aaclayton (approved):
Good!

PR #778, opened 2026-04-01
Fix `prepareMovement`-using hooks

### LINKED ISSUE #781: Bug: True Grit, Seasoned Veteran, & Hold Fast no longer increase engagement, Distance Runner no longer increases stride
> True Grit I intended, but shredded I was

Looks like a regression was introduced by the actor preparation refactor in [efce48feb393505d8cf90934e9372a035be7a710](https://github.com/foundryvtt/crucible/commit/efce48feb393505d8cf90934e9372a035be7a710) (`Reorganize actor data preparation to allow for ability scores to be targeted by actor hooks or active effects`).

`True Grit` is supposed to increase an actor's Engagement by `+1`, but that bonus no longer sticks. The talent hook still adds directly to `movement.engagement`, but the refactor now recalculates final engagement afterward from `baseEngagement + engagementBonus`, which overwrites the earlier direct write.

## What's happening

Before the refactor, this kind of hook mutation worked:

```js
HOOKS.truegrit00000000 = {
  prepareMovement(_item, movement) {
    movement.engagement += 1;
  }
};
```

After the refactor, movement prep now does this:

```js
this.parent.callActorHooks("prepareMovement", this.movement);
this._prepareFinalMovement();
```

And `_prepareFinalMovement()` rebuilds engagement here:

```js
const {mainhand, offhand} = this.parent.equipment.weapons;
if ( mainhand && mainhand.system.properties.has("engaging") ) m.engagementBonus += 1;
if ( offhand && offhand.system.properties.has("engaging") ) m.engagementBonus += 1;
m.engagement = m.baseEngagement + m.engagementBonus;
```

So the hook writes to `movement.engagement`, then final prep immediately replaces that value.

roth-michael:
They shouldn't be trying to modify the final value, but the bonus value, for whatever field.

### DIFF
--- a/module/hooks/talent.mjs
+++ b/module/hooks/talent.mjs
@@ -267,7 +267,7 @@ HOOKS.demolitionist000 = {
 
 HOOKS.distancerunner00 = {
   prepareMovement(_item, movement) {
-    movement.stride += 1;
+    movement.strideBonus += 1;
   }
 };
 
@@ -317,7 +317,7 @@ HOOKS.healer0000000000 = {
 
 HOOKS.holdfast00000000 = {
   prepareMovement(item, movement) {
-    if ( this.equipment.weapons.shield ) movement.engagement += 1;
+    if ( this.equipment.weapons.shield ) movement.engagementBonus += 1;
   }
 };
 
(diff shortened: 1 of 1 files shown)


aaclayton (approved):
Good fixes, I think these hooks predated the refactor where I added targetable bonus fields for each of these.

PR #779, opened 2026-04-03
[774] Consider adding group test hotkey with custom dialog for fast on the …

Plushtoast:
here is an example of the quick group check selection window.
let me know if you want it, otherwise i ll add it to crucibletongs module

* adds hotkey g to open quick select. quick select just shows all available skills and then opens the group check dialog
* moves hotkey config into its own file (more hotkeys in the future for ember speed runs)
* adds new category for crucible classes called "tools"
* adds a reusable grid class (i think there was none) called .asbestos. crucible must be fireproof and the shent build blue crystal things everywhere in ancient times so some amount of asbestos in crucible is logical

Fixes #774

### DIFF
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -22,6 +22,7 @@ import * as hooks from "./module/hooks/_module.mjs";
 import {handleSocketEvent} from "./module/socket.mjs";
 import {registerEnrichers} from "./module/enrichers.mjs";
 import * as chat from "./module/chat.mjs";
+import {registerKeybindings} from "./module/hotkeys.mjs";
 import * as interaction from "./module/interaction.mjs";
 import Enum from "./module/const/enum.mjs";
 import CrucibleTalentNode from "./module/const/talent-node.mjs";
@@ -395,13 +396,7 @@ Hooks.once("init", async function() {
   });
 
   // Register keybindings
-  game.keybindings.register("crucible", "confirm", {
-    name: "KEYBINDINGS.ConfirmAction",
-    hint: "KEYBINDINGS.ConfirmActionHint",
-    editable: [{key: "KeyX"}],
-    restricted: true,
-    onDown: chat.onKeyboardConfirmAction
-  });
+  registerKeybindings();
 
   // Patch door sound radius - can we do this better elsewhere?
   Object.defineProperty(foundry.canvas.placeables.Wall.prototype, "soundRadius", {
--- a/lang/en.json
+++ b/lang/en.json
@@ -1275,6 +1275,7 @@
       "Title": "Group Check",
       "ChatTitle": "{skill} Check (DC {dc})",
       "DialogTitle": "{skill} Check: {name}",
+      "QuickSelection": "Group Check Quick Selection",
       "NoUserForActor": "No active user found for: {name}",
       "STATUSES": {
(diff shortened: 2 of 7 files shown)


Plushtoast:
grafik

aaclayton (changes requested):
I probably need to think about this one for a little bit. Things I like about this PR:

* Moving keybindings to a separate file is smart, there will be more in the future.
* There definitely needs to be *some* way to start a group check without having to write a journal text enricher to do it, so I am aligned on the need for something here.

Things I'm not so sure about

* The approach of a dialog that leads to a dialog, it feels to me like the step of configuring the group check could somehow be more integrated with the group check UI.
* I am envisioning that in the future group checks will need to allow each rolling user a limited range of choice about which skill to use. Consider a group check where you can do either diplomacy OR deception OR intimidation and each person in the group can choose which they prefer. We might want to implement that underlying framework first before we decide on what exactly is the UI/UX to initiate it.

Things I don't like:

* Using a CSS class name like `asbestos` - too easy to forget what it does. We should prefer brutalist and descriptive CSS classes like `icon-grid` or `button-grid` or something.
* No singleton handling or deduplication. Each press of G opens a new window.
* We should name the file `keybindings` not `hotkeys`.

Next steps:

1. Maybe we could filter out the things that are agreed upon for now and implement them for example factor out hotkeys and add a hotkey to start a group check. 
2. I would suggest making the UI for that "start a group check" intentionally minimalist for now because I think it will need to change the future and I don't think we should overdesign now until we know what our long term needs are.
3. File the rest of the ideas as github issues we can track for later work

Plushtoast:
ok i split it up. i agree that dialog into dialog is not nice, but was the fastest way to do it:

potentially better would be if the standard dialog can act like a wizard/state machine:
* start standard dialog in wizard mode
* dialog shows all skills in asbestos... *cough* i mean crucible-grid 
* user selects skill
* dialog rerenders (and has button to check skill)
* skill select can allow multiple selects
* query then forwarded to player
* player also sees wizard but this time skill select if skill multi selection is larger than 1 (and with skill pips)
* player selects skill
* normal flow
* chatcard must reflect skill choice and skill options

agrred on would be then, right?
* a keybindings file.
* crucible-grid (to not collide with modules) css class in theme

the hotkey wont work though because how would you select a skill to roll on then without dialog->dialog?

i could also instead upgrade the skill dialog so it has a selector for boons/banes dc, users then immediately runs group check without standard dialog in between. is that the minimalist approach or do you prefer the state machine like above?

aaclayton:
> potentially better would be if the standard dialog can act like a wizard/state machine:

I agree we want to move towards this sort of multi-step dialog - I'm not sure it is the right *next thing* to make.

> agrred on would be then, right?
> * a keybindings file.
> * crucible-grid (to not collide with modules) css class in theme
> the hotkey wont work though because how would you select a skill to roll on then without dialog->dialog?

We can have a Dialog (choose which skill) -> Dialog (do the check) workflow for now, but if we know it's temporary we should make the first dialog (choose which skill) as basic as possible so it's just a temporary placeholder.

aaclayton:
Closed in favor of some other new features for multi-skill group checks and a convenience macro for starting the process.

PR #817, opened 2026-04-08
[#765] Special handling to reverse `system.loaded` changes on strike/reload

### LINKED ISSUE #765: Undoing a strike with a mechanical weapon doesn't bring it back to a loaded state

roth-michael:
Closes #765 
Worth considering a more generic way of handling reversion of changes like this, but this feels like a solid solution at least for the time being.

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -686,6 +686,12 @@ export const TAGS = {
           this.usage.actorUpdates.items.push({_id: w.id, "system.loaded": false});
         }
       }
+    },
+    async confirm(reverse) {
+      if ( !reverse ) return;
+      for ( const itemUpdate of (this.selfEvents.actorUpdate.actorUpdates.items ?? []) ) {
+        if ( itemUpdate.system?.loaded === false ) itemUpdate.system.loaded = true;
+      }
     }
   },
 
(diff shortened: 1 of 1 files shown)


roth-michael:
Needs a refactor with the new event system. Can maybe now support the more generic version simply by restoring the weapon snapshot on revert.

module/const/action.mjs
```diff
@@ -913,6 +919,12 @@ export const TAGS = {
       if ( this.usage.weapon ) {
         this.usage.actorUpdates.items.push({_id: this.usage.weapon.id, "system.loaded": true});
       }
+    },
+    async confirm(reverse) {
+      if ( !reverse ) return;
+      for ( const itemUpdate of (this.selfEvents.actorUpdate.actorUpdates.items ?? []) ) {
+        if ( itemUpdate.system?.loaded ) itemUpdate.system.loaded = false;
+      }
     }
```
aaclayton, line 928: I think a robust way to do this in an events stream context would be this:

1. Identify the weapon and record actorUpdates in preActivate
2. Snapshot the current weapon state in postActivate
3. Revert to snapshotted weapon state in confirm(reverse)

```suggestion
  preActivate() {
    this.usage.actorUpdates.items ||= [];
    this.usage.weapon ??= this.getValidWeaponChoices()[0]?.item;
    if ( this.usage.weapon ) this.usage.actorUpdates.items.push({_id: this.usage.weapon.id, "system.loaded": true});
  },
  postActivate() {
    const activation = this.selfEvents?.activation;
    if ( activation && this.usage.weapon ) {
      activation.weapon = this.usage.weapon.system.snapshot();
    }
  },
  async confirm(reverse) {
    if ( !reverse ) return;
    const {activation, actorUpdate: updateEvent} = this.selfEvents;
    const snapshot = activation?.weapon;
    if ( !snapshot || !updateEvent ) return;
    const items = updateEvent.actorUpdates.items || [];
    const idx = items.findIndex(w => w._id === snapshot.id);
    if ( idx === -1 ) return;
    const {id: _id, ...system} = snapshot;
    items[idx] = {_id, system};
  }
```
It's a little longer, but more robust because it handles other stateful properties of the weapon so we don't rely on assuming that reloaded state is the only thing that changed in the action that bears this `reload` tag.

I didn't test this, so there might be a bug or two in the exact syntax, but hopefully you get the gist. Give it a try?

aaclayton (changes requested):

PR #818, opened 2026-04-08
[#762] Compare almost all item fields when determining whether some item can stack on an existing item

### LINKED ISSUE #762: Ensure we only stack stackable items when their quality & enchantment (& any other relevant fields) match
Currently, purely checking the `stackable` property of both and their identifiers.

roth-michael:
Closes #762 
`CrucibleItem#isStackableWith` can receive refinement as needed. Currently will definitely not have any false positives (erring on the side of being over restrictive).

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -410,6 +410,29 @@ export default class CrucibleItem extends foundry.documents.Item {
   }
 
   /* -------------------------------------------- */
+
+  /**
+   * Test whether an item is stackable with another item
+   * @param {CrucibleItem} other
+   * @returns {boolean}
+   */
+  isStackableWith(other) {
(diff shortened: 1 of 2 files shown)


module/documents/item.mjs
```diff
@@ -409,6 +409,29 @@ export default class CrucibleItem extends foundry.documents.Item {
...
+    const cleanOther = other.toObject();
+    for ( const data of [cleanData, cleanOther] ) {
+      delete data._id;
+      delete data._stats;
+      delete data.system.quantity;
+      delete data.sort;
+      delete data.ownership;
+      delete data.folder;
+    }
+    return foundry.utils.equals(cleanData, cleanOther);
```
aaclayton, line 432: I guess anything with an affix shouldn't stack, so we might consider early returning false if either of the items have `effects.size`?

aaclayton (changes requested):

PR #839, opened 2026-04-10
[#838] Fix validity criteria for weapon choices

### LINKED ISSUE #838: Wielding both a melee and ranged weapon does not present a choice when striking

roth-michael:
Closes #838 
Also added some visual feedback in the action use dialog header (and resulting chat card) for which weapon is actually being reloaded.

Changes to checking weapon validity for a given action:
- When determining melee/ranged, check _source_ tags as well as current. This ensures something like Strike (which would have its `ranged` tag or `melee` tag removed depending on the currently-selected weapon) gets proper consideration for both types of weapons
- Return `available` and `eligible` instead of just `isValid`. Weapons for whom `eligible` is `false` aren't returned, even if `strict` is `false`. For example: A melee weapon will _never_ be in the list of valid weapons for Precise Shot, even with `isValid: false`

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -907,9 +907,16 @@ export const TAGS = {
         throw new Error(_loc("ACTION.WARNINGS.NoReloadRequired"));
       }
     },
+    prepare() {
+      this.usage.weapon ??= this.getValidWeaponChoices()[0]?.item;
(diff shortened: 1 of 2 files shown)


aaclayton (approved):
Merging! I have a few tidying things I'd like to do but I'll amend.

PR #846, opened 2026-04-12
Fix ember hook patches

roth-michael:
Issues fixed:
- `const ember = globalThis.ember` is no good when called outside of a hook, ember hasn't yet registered itself. So made it a `let` and assigned it prior to applying patches (and renamed the existing `ember` const in there to `emberModule` for clarity).
- Any any effect changes modified `mode` -> `type` (didn't break anything, but why not).
- Crystalize Wounds and Extreme Metabolism were recording activation events, but that's a no-go since they're already defined and are singletons. So modify existing instead.
- Anything adding changes should be targeting an effect's `system.changes`, not just `changes`.
- Ember Blaze should be `postActivate` otherwise we lose knowledge of "what were your resource values before you used this?"
- Magnetic Disarm should be `postActivate`, and simplified the filtering for success.

### DIFF
--- a/module/hooks/ember.mjs
+++ b/module/hooks/ember.mjs
@@ -10,7 +10,7 @@ const EMBER_050 = {
     crystalizeWounds: {
       preActivate() {
         const health = this.actor.level * 2;
-        this.selfEvents.activation.resources = [{resource: "health", delta: health}];
+        this.selfEvents.activation.resources.push({resource: "health", delta: health});
         const res = this.actor.abilities.toughness.value;
         Object.assign(this.effects[0].system, {
           changes: [
@@ -23,7 +23,7 @@ const EMBER_050 = {
     },
     extremeMetabolism: {
       preActivate() {
-        this.selfEvents.activation.resources = [{resource: "action", delta: 1}];
+        this.selfEvents.activation.resources.push({resource: "action", delta: 1});
         const dodgeBonus = Math.ceil(this.actor.abilities.dexterity.value / 2);
         Object.assign(this.effects[0].system, {
           changes: [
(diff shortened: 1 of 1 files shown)


module/hooks/ember.mjs
```diff
@@ -3,33 +3,33 @@
...
-const ember = globalThis.ember;
+let ember;
 
 const EMBER_050 = {
   action: {
     crystalizeWounds: {
       preActivate() {
         const health = this.actor.level * 2;
-        this.recordEvent({type: "activation", resources: [{resource: "health", delta: health}]});
+        this.selfEvents.activation.resources = [{resource: "health", delta: health}];
```
aaclayton: We shouldn't assume that the activation event doesn't have any resources already incurred, we should push to an existing array rather than redefining the array.

module/hooks/ember.mjs
```diff
@@ -182,10 +184,11 @@ const EMBER_PATCHES = {
  * Called during the "setup" Foundry hook, after Ember has registered its hooks in "init".
  */
 export function applyEmberPatches() {
-  const ember = game.modules.get("ember");
-  if ( !ember?.active ) return;
+  ember = globalThis.ember;
+  const emberModule = game.modules.get("ember");
+  if ( !emberModule?.active ) return;
```
aaclayton: `globalThis.ember` and `game.modules.get("ember")` are the same object, are they not?

aaclayton (changes requested):
Thanks for these fixes, two important callouts, but these are good fixes.

module/hooks/ember.mjs
roth-michael: Indeed they are. Whoops.

aaclayton (approved):

PR #848, opened 2026-04-12
Revert to previous canvas layer after closing region-using action use dialog

roth-michael:
Can only revert if we don't need the preview, otherwise the preview gets destroyed. I think it might make sense to have this "preview" be an actual created region that we just explicitly clean up for if need be. But for now this is a simple, clean solution.


aaclayton:
Closed in favor of an alternative solution - although we still need to resolve an issue with losing token control.

PR #862, opened 2026-04-13
Mark non-talisman weapons ineligible for talisman-tag actions, improve talisman tag's `canUse`

roth-michael:
Changes:
1. `talisman` tag? Only valid weapons are `talisman1` or `talisman2` category
2. Remove `canUse` from refocus, instead improving the `talisman` tag's `canUse`:
3. `talisman` tag's `canUse` returns false if no equipped talisman, preventing it from being shown in favorites. I _think_ we can probably remove the remaining `this.usage.strikes` check, due to the implementation of (1) it shouldn't be possible to end up in a state where a non-talisman is being used, but left it just in case
4. Removed refocus's `preActivate`, as it's rendered useless by (1) and (3)

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -312,9 +312,9 @@ export const TAGS = {
     canUse() {
       const {mainhand: mh, offhand: oh} = this.actor.equipment.weapons;
       const categories = ["talisman1", "talisman2"];
-      if ( !categories.includes(mh?.category) && !categories.includes(oh?.category) ) return false;
-      if ( !this.usage.strikes.every(w => w.config.category.training.includes("talisman")) ) {
-        throw new Error(_loc("ACTION.WARNINGS.RequiresTalisman"));
+      if ( (!categories.includes(mh?.category) && !categories.includes(oh?.category))
+        || !this.usage.strikes.every(w => w.config.category.training.includes("talisman")) ) {
+        throw new Error(_loc("ACTION.WARNINGS.RequiresTalisman", {action: this.name}));
       }
     }
   },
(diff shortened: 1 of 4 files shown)


module/const/action.mjs
```diff
@@ -310,6 +310,9 @@ export const TAGS = {
     tooltip: "ACTION.TAG.TalismanTooltip",
     category: "requirements",
     canUse() {
+      const {mainhand: mh, offhand: oh} = this.actor.equipment.weapons;
+      const categories = ["talisman1", "talisman2"];
+      if ( !categories.includes(mh?.category) && !categories.includes(oh?.category) ) return false;
```
aaclayton: Remind me about the semantic difference between the soft `return false` vs the explicit thrown error with informative message. Wouldn't the explicit `RequiresTalisman` error string satisfy both use cases in a more interpretable way than just a generic "You can't use this action" message would that we get from `return false`?

aaclayton (approved):
Approved with concern about `return false` in `canUse` instead of the more explicit error string

module/const/action.mjs
roth-michael: Yeah, it would. No semantic difference:
```js
  _displayOnSheet() {
    try {
      for ( const test of this._tests() ) {
        if ( !(test.canUse instanceof Function) ) continue;
        if ( test.canUse.call(this) === false ) return false;
      }
    } catch(err) {
      return false;
    }
    return true;
  }
  ```
I'll have it so that either failure throws the same error.

aaclayton (approved):

PR #863, opened 2026-04-13
Update `CrucibleActor#_cachedResources` only after max resources have been prepared

roth-michael:
Calling in constructor was too early, as resources aren't prepared and therefore only have `value`s. Moved into `prepareDerivedData`, which meant changing `CrucibleActor##updateCachedResources` from a hard private to an internal `CrucibleActor#_updateCachedResources` (because of synthetic actors... I think? For some reason despite being `CrucibleActor`s, they were throwing type errors about the receiver not being an instance of class `CrucibleActor`).

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -22,10 +22,6 @@ const {DialogV2} = foundry.applications.api;
  * The Actor document subclass in the Crucible system which extends the behavior of the base Actor class.
  */
 export default class CrucibleActor extends Actor {
-  constructor(...args) {
-    super(...args);
-    this.#updateCachedResources();
-  }
 
   /**
    * The Actions which this Actor has available to use.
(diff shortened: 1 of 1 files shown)


module/documents/actor.mjs
```diff
@@ -2854,8 +2858,9 @@ export default class CrucibleActor extends Actor {
 
   /**
    * Update cached resources for this Actor.
+   * @internal
    */
-  #updateCachedResources() {
+  _updateCachedResources() {
     this._cachedResources ||= {};
```
aaclayton, line 2866: If this no longer happens at constructor time, I think it's impossible for this to be undefined?

```suggestion
```

module/documents/actor.mjs
```diff
@@ -2854,8 +2858,9 @@ export default class CrucibleActor extends Actor {
 
   /**
    * Update cached resources for this Actor.
+   * @internal
    */
-  #updateCachedResources() {
+  _updateCachedResources() {
```
aaclayton: We could consider a signature here which internalizes the correct logical test rather than needing the caller to know it:

```suggestion
  _updateCachedResources(force=true) {
    if ( !force && foundry.utils.isEmpty(this._cachedResources) ) return;  
```

aaclayton (approved):
Approved, but two minor comments you could clean up please?

module/documents/actor.mjs
roth-michael: It is, oddly. It'll be undefined for synthetic actors it seems.

aaclayton (approved):

PR #873, opened 2026-04-14
Loot enrichers: allow `broken=true` token, ensure text wrapping when in a crucible chat message

roth-michael:
Future consideration: Display "(Broken)" suffix or "Broken" prefix in enricher, maybe. Already visible in item header in tags, but would be useful to be able to tell from an enricher at a glance.


aaclayton (approved):
Thanks!

Issue #611, opened 2026-01-23
V14 Aura Behaviors


aaclayton:
Dependent on #609 and #610

- [ ] Implement a region behavior type for Aura actions that directs region events back to the originating Action for handling via action hooks.
- [ ] Track as part of the Aura behavior a reference to the ActiveEffect which controls the Aura.

Issue #610, opened 2026-01-23
V14 Aura Regions


aaclayton:
- [ ] Enable aura-type Actions to provide `Region` documents that are linked to the originating `Token`. 
- [ ] Record the existence of the aura as part of `ActiveEffect` system data.
- [ ] Automatically disable the aura if the associated `ActiveEffect` is disabled or suppressed.
- [ ] Automatically remove the aura when the associated `ActiveEffect` is deleted.

roth-michael:
> Automatically disable the aura if the associated `ActiveEffect` is disabled or suppressed

There are some ways we could attempt to handle this currently, but I think the _right_ way may require a core change: https://github.com/foundryvtt/foundryvtt/issues/14496

At best right now we can enable/disable all behaviors and change visibility, but since that's effectively what `hidden` does, I'd prefer to be able to leverage that.

PR #890, opened 2026-04-19
Add aura target type

roth-michael:
Happy to build on this PR or to split up the aura work into multiple small PRs.
Addresses the first bullet of #610 
Next PR would be the remaining bullets.
Final PR would be some version of #611


aaclayton (approved):
Looks like a good starting point and safe to merge in before reviewing other work.

PR #912, opened 2026-04-24
Russian translation

danielrab:
The initial commit just includes the setup and the translation of the attribute names as a way to check that the setup is correct.


danielrab:
Closing in favor of https://github.com/foundryvtt/crucible/pull/916 as I don't see a reason for me to continue working on this when there's a finished translation.

PR #917, opened 2026-04-27
Add Polish translate

Lioheart:


aaclayton:
Thank you @Lioheart, can you please confirm that you (1) are a fluent Polish speaker and (2) have personally reviewed these translation strings?

Lioheart:
@aaclayton I was born in Poland and have checked these lines twice. I'd have to play through it to find any additional errors, but right now I'm 95% sure it's an accurate translation

PR #920, opened 2026-04-28
add DSN integration for damage types

### LINKED ISSUE #919: Add DSN damage types integration
The latest update to the Dice So Nice module (released a few hours ago) introduced a way to customize the dice depending on the damage type.
All we need to do on our end to take advantage of this is to set either `DiceTerm.options.type` on each of the terms or set `Roll.options.type` on the roll as a whole.
I also propose we set that field on skill checks to the skill used in the roll, which would allow users to customize the dice used for each skill check.

danielrab:
closes https://github.com/foundryvtt/crucible/issues/919

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -903,6 +903,7 @@ export const TAGS = {
         skill: bonuses.skill ?? 0,
         enchantment: bonuses.enchantment,
         defenseType,
+        damageType,
         dc: target.defenses[defenseType].total
       });
       await roll.evaluate();
--- a/module/dice/standard-check.mjs
+++ b/module/dice/standard-check.mjs
@@ -58,6 +58,7 @@ export default class StandardCheck extends Roll {
     }
     super(formula, data);
     this.actor = game.actors.get(this.data.actorId);
+    this.options.type = this.data.damage?.type ?? this.data.type; // Integration with DSN
   }
 
   /* -------------------------------------------- */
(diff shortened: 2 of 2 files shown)


danielrab:
upon further testing, one downside for the implementation I provided seems to be that the damage type only applies when the hit is actually successful. idk if I like that.

roth-michael:
Not sure what the ideal way to handle that oddness is. `damageType` is available in the input `data`. Could do something like:
```diff
- constructor(formula, data) {
+ constructor(formula, data, options) {
    if ( typeof formula === "object" ) {
      data = formula;
      formula = ""; // Replaced later
    }
-   super(formula, data);
+   super(formula, data, options);
    this.actor = game.actors.get(this.data.actorId);
+   this.options.type ??= data.damageType ?? this.data.damage?.type ?? this.data.type; // Integration with DSN
  }
```
(the `options` pass is something we should be doing anyway I'm pretty sure). There may be a cleaner way to do this _just_ inside `AttackRoll`.

aaclayton:
Agree with @roth-michael comments here @danielrab, you need to configure this roll type in a way that does not depend on the attack being successful. The style of dice used should be based on the tags and features of the action attempted, not dependent on the roll result.

danielrab:
I honestly can't figure out what comes from where, but there's no `damageType` on `data`, what there is however is a consistently passed third argument with `{type: "bludgeoning"}` for example. Passing it instead of `data` as @roth-michael 's suggested diff does, breaks things, which is to be expected with it only having a single field.

aaclayton (approved):

PR #921, opened 2026-04-28
Improve Group Skill check rendering using Dice So Nice

JiDW:
This change improves the UI feels during a group skill check by only hiding the relevant Roll result and keeping the name of the actor visible instead of hiding the entire line and creating a more jarring UI shift.  

Only works with DsN v6+. Will not break older versions.  

image

image


aaclayton (approved):
Thanks @JiDW!

PR #922, opened 2026-04-28
[#902] Enable member actor drag and drop for group actors

### LINKED ISSUE #902: Enable initiating an Actor drag operation from the Party sheet to transfer actors or place them into an area map

Aioros:
Closes #902

### DIFF
--- a/module/applications/sheets/actor-group-sheet.mjs
+++ b/module/applications/sheets/actor-group-sheet.mjs
@@ -111,7 +111,7 @@ export default class CrucibleGroupActorSheet extends api.HandlebarsApplicationMi
   async _onDragStart(event) {
     const { actorId } = event.currentTarget.dataset;
     if ( actorId ) {
-      const actor = this.actor.system.members.actors.find(a => a.id === actorId);
+      const actor = this.actor.system.actors.find(a => a.id === actorId);
       if ( actor ) {
         const dragData = { origin: this.id, ...actor.toDragData() };
         event.dataTransfer.setData("text/plain", JSON.stringify(dragData));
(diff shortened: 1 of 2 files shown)


module/applications/sheets/actor-group-sheet.mjs
```diff
@@ -107,8 +107,23 @@ export default class CrucibleGroupActorSheet extends api.HandlebarsApplicationMi
   /*  Drag and Drop                               */
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _onDragStart(event) {
+    const { actorId } = event.currentTarget.dataset;
+    if ( actorId ) {
+      const actor = this.actor.system.members.actors.find(a => a.id === actorId);
```
roth-michael:
```suggestion
      const actor = this.actor.system.actors.find(a => a.id === actorId);
```
We've got `actors` which populates in derived data prep to be the Actor of anyone in `members` that is a valid world actor.

module/applications/sheets/actor-group-sheet.mjs
```diff
@@ -107,8 +107,23 @@ export default class CrucibleGroupActorSheet extends api.HandlebarsApplicationMi
...
+        const dragData = { origin: this.id, ...actor.toDragData() };
+        event.dataTransfer.setData("text/plain", JSON.stringify(dragData));
+      }
+    }
+    return super._onDragStart(event);
+  }
+
   /** @override */
   async _onDropActor(event, actor) {
+    const data = CONFIG.ux.TextEditor.implementation.getDragEventData(event);
```
roth-michael:
```suggestion
    const data = CONFIG.ux.TextEditor.getDragEventData(event);
```
`CONFIG.ux.TextEditor` will already be the `implementation`, so no need to double up there.

roth-michael (approved):
Couple minor improvements but looks good to me.

aaclayton (approved):
Thanks @Aioros! Thanks for the review @roth-michael

Issue #924, opened 2026-04-29
The returning affix prevents the item from being dropped but not unequipped


OsiJr:
I have a player using a chain hook; it _feels_ buggy that it gets unequipped after each `throw`.

PR #925, opened 2026-04-29
Fix for returning so that the item stays equipped as well

OsiJr:
This is the fix I've been using locally for #924

### DIFF
--- a/module/hooks/affix.mjs
+++ b/module/hooks/affix.mjs
@@ -141,8 +141,11 @@ HOOKS.returning = {
   preActivateAction(item, action) {
     const updates = action.selfUpdateEvent.actorUpdates.items;
     if ( !updates ) return;
-    const update = updates.find(u => (u._id === item.id) && u.system?.dropped);
-    if ( update ) delete update.system.dropped;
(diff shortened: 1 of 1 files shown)


roth-michael:
Would this not also prevent deliberately un-equipping during combat?

OsiJr:
Yes it would. Isn't that fun. 😅

roth-michael:
I think keeping the original criteria (matching id, update has dropped) and just putting in `equipped: true` as part of the update would suffice. 

Or rather, deleting both the `dropped` and `equipped` updates on such a match.

aaclayton:
I think we want to:

1. Verify that the action is a `strike`. There could be other circumstances in which the weapon would be dropped, but the Returning affix should specifically avoid dropping *when you perform a strike*. Current logic doesn't support this, but something we should add. If not a strike, bail out.
2. Find an item update where the weapon is becoming dropped. As long as it is currently equipped, it stays equipped. As long as it is currently not dropped, it remains not dropped.

Make sense?

aaclayton:
Resolved in https://github.com/foundryvtt/crucible/commit/b0eb45187e3d95624c2bab59aa3d4ce3a8ae4049

PR #927, opened 2026-05-01
[#918] Various improvements to action use dialog flow

### LINKED ISSUE #918: Action Use Dialog improvements
- If the region required by an action is self-anchored and not rotatable (pulse, aura), auto-place the preview without requiring any "place region" button
- If targets are invalid, disable the Use button
- Change behavior of Use button such that it first prompts region placement/movement planning if required, and if that results in a failure simply un-minimize and don't actually proceed

roth-michael:
Closes #918 
Changes are as described in the issue. 
- Self-anchored circles or emanations do not generate a "place region" button, and are instead auto-placed upon render.
- Invalid targets, or 0 targets with a `single`-target-type action, disable the "Use" button with a generic tooltip.
- The "Use" button is no longer disabled when a region needs placing or a movement needs planning - instead, said placement/planning will be prompted upon clicking "Use" - a cancelled placement/plan will not continue the submit, otherwise strict target acquisition is performed to ensure valid targets; if not, won't continue the submit. Otherwise, submit as normal, then close (had to make it `closeOnSubmit: false` for this).
- Only call `ActionUseDialog#_clearTargetRegion` when changing spellcraft components if the `gesture` is being changed (nothing else should change the shape of the region). Also, within `ActionUseDialog#_clearRegionTarget`, only reset targets if there are existing region targets (i.e. Ray to Pulse, vs say Touch to Influence).

This last solution isn't _perfect_, as we may want to visually reset targets when switching from, for instance, Touch to Ray, but since it's impossible to proceed without placing a Ray template at that point I don't think it's a big enough concern to outweigh the benefit of being able to switch damage type or inflection without having to retarget. Another potential thing to address with it is re-acquiring targets on inflection change (for instance, an Earthen Ray vs a Shaped Earthen Ray, where the latter should now exclude allies). At time of cast, targets will be correct, but while the dialog is up if an inflection changes valid targets it won't be represented, and if ever we have an inflection that changes the actual _shape_ of a gesture, we'll need to revisit this.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -427,6 +427,7 @@
       "DirectMovementOnly": "This action requires a direct movement path with no intermediate waypoints.",
       "IncorrectTargets": "You may only designate up to {number} {type} targets for {action}!",
       "InvalidTarget": "You must designate at least one {type} target for {action}!",
+      "InvalidTargetsGeneric": "Invalid targets for this action!",
       "LastCriticallyMissed": "You may only perform {action} after a basic Strike which did not critically miss.",
       "MaximumRange": "A selected target is further than the maximum range of {max} feet.",
       "MinimumRange": "A selected target is closer than the minimum range of {min} feet.",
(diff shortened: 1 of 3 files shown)


aaclayton (approved):
Thanks @roth-michael good looking improvements

PR #991, opened 2026-05-07
[#990] Ensure `_source` contains items when "creating" new Hero at the end of creation workflow

### LINKED ISSUE #990: A non-GM attempting to finalize character creation throws an error and blocks the update
More specifically, when creating embedded documents via an `update` of the parent actor, rather than using `createEmbeddedDocuments`, because `_state.source` does not exist (since the to-be-created items aren't yet created on the actor), and so the following line in `SchemaField#_cleanType` evaluates to `true`:
```js
const cleanField = (name in data) || !options.partial || !_state.source;
```
which means that the `ownership` field is "cleaned" and therefore added to the update data for each item.

With https://github.com/foundryvtt/foundryvtt/issues/5241 being an open issue, it seems that maybe creating embedded documents as part of an `update` operation isn't a _fully_ supported pattern, so I'm not certain whether this is a "Core issue, it shouldn't be cleaning these fields" problem or a "Crucible issue, we should be batching the actor update & embedded collection update rather than performing both with an `update`" problem.

roth-michael:
Closes #990 
Two-parter:
1. `updateSource` with `items` so that they don't erroneously get their `ownership` field added during data cleaning
2. Give equipment a random ID, rather than deleting, so that when the `sourceIdMap` is created during cleaning, it actually has values for the equipment

As mentioned in the issue, there's potentially a Core issue to be fixed here.

### DIFF
--- a/module/applications/sheets/hero-creation-sheet.mjs
+++ b/module/applications/sheets/hero-creation-sheet.mjs
@@ -1112,6 +1112,8 @@ export default class CrucibleHeroCreationSheet extends HandlebarsApplicationMixi
 
     // Update the actor and render the regular sheet
     this.#complete = true;
+    // TODO: Determine whether this is the ideal fix, we should do something different, or core needs a fix here
+    this.document.updateSource({items: creationData.items}, creationOptions);
     await this.document.update(creationData, creationOptions);
     await this.close({dialog: false});
   }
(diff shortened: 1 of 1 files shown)


aaclayton:
Basically ended up adopting your approach, but I re-engineered it because I took the time to understand the upstream bug and propose an upstream fix for next v14 release. Closed via https://github.com/foundryvtt/crucible/commit/85921facef3c9a438091a53d90c87135ca3e53a9

PR #1008, opened 2026-05-10
Add "Crucible Action" region behavior, use for auras/walls & expose to end user

### LINKED ISSUE #611: V14 Aura Behaviors

roth-michael:
Closes #611 
Addresses most of #610 (still not the "auto disable if AE suppressed or disabled" bit)

Adds new Region Behavior for Persistent Areas of Effect. This behavior can have configured on it:
- The "origin" action identifier & actor (used to determine whether a given token should trigger an event, based on whether it could be targeted by the original action)
- An action to be "remotely" performed by the origin actor on a triggering token
- Whether to fire on enter, on turn start, or both
- Whether to only apply once per round (currently nonfunctional)

On event trigger, will check whether the origin action would target the triggering token. If so, will check if once/round criterion allows for the trigger. If so, will programmatically execute the configured Action with the origin actor as the action "user," and the triggering token as the action target.

In the case of pre-placed scene regions, the base actor & action would correspond to some actor to be the dedicated "attacker" and, optionally, a dedicated action used for target validation; if no origin action, assume that any triggering token should be targeted.

In the case of a persistent region created by an action, each action now tracks `regionBehavior`, which holds a pared down version of the actual behavior's data: A `name`, and `system`, which holds `events` (when to trigger), `oncePerRound`, and `actionToPerform`, which itself holds `id`, `name`, `img`, `description`, `effects`, and `tags`. Then, when originally _creating_ the region upon action use, a behavior with that data stored in it is created as well (in a disabled state). Upon confirmation, the behavior is enabled.

One can edit the above `regionBehavior` data on an Action, whether owned by an actor or not. This is done by creating a transient RegionBehavior in memory only and opening its sheet, then redirecting submissions to modify the parent action, rather than the (nonexistent) document itself. In the case of an action on an actor-owned item, `system.actor` and `system.actionIdentifier` are automatically set and used to determine the parent item in need of update. In the case of an action on a parentless item, we use the `flags.crucible.itemUuid` which is created on the transient RegionBehavior.

TODOs:
- Style pass (visual & code)
- ~~Figure out why an update to `sytsem.affectedActors` isn't working. Possible core bug~~


aaclayton (changes requested):
@roth-michael when you get a chance could you rebase this on current master. A few things in particular I'd like to look at more closely:

1. Clean-up on removal of active effect - I assume we need some wiring so that when an active effect that is responsible for managing a persistent region is removed, that region is also removed? Does this exist and I missed it?
2. Work on your TODO about `affectedActors`, maybe consider whether we want/need to `await action.use()`.
3. Perhaps we should limit the select of available events similar to the `behavior-hazard?
4. If not in time for the next patch, maybe you should split out and get merged the fixes for the `generic` tag to address enchantment bonus and restoration usage?

roth-michael:
Sure I’ll take a run at this tomorrow afternoon.

1. This is already coded in, just needed to add any regions to `system.regions`
2. I’ll check and see if maybe the TOF bug has been inadvertently addressed (or maybe figure out that it's a "me" problem). https://github.com/foundryvtt/foundryvtt/issues/14367 is the relevant issue. 
3. Good thinking, makes sense. 
4. Think Kim included one of the `generic` fixes in the plummet implementation, but I can pull out the restoration one if need be.

crucible.mjs
```diff
@@ -192,10 +192,12 @@ Hooks.once("init", async function() {
 
   // Region Behavior document configuration
   Object.assign(CONFIG.RegionBehavior.dataModels, {
-    "crucible.hazard": models.CrucibleHazardRegionBehavior
+    "crucible.hazard": models.CrucibleHazardRegionBehavior,
+    "crucible.persistentAOE": models.CruciblePersistentAOERegionBehavior
```
aaclayton: Sorry to not call this out substantially earlier, but I don't actually like `crucible.persistentAOE` as a type ID:

1. All behaviors are persistent, so that part seems redundant
2. All regions are areas of effect, so that part seems redundant

I think we need to scrutinize *what is this behavior responsible for doing. Region behaviors are designed to be inherently modular, you mix-and-match the behaviors you need, so I think we need clarity on "what exactly are we mixing in here?"

module/models/behavior-aoe.mjs
```diff
@@ -0,0 +1,68 @@
...
+  static LOCALIZATION_PREFIXES = ["REGION_BEHAVIORS.PERSISTENT_AOE"];
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    const {TOKEN_ENTER, TOKEN_EXIT, TOKEN_MOVE_IN, TOKEN_MOVE_OUT, TOKEN_MOVE_WITHIN,
+      TOKEN_TURN_START, TOKEN_TURN_END, TOKEN_ROUND_START, TOKEN_ROUND_END} = CONST.REGION_EVENTS;
+    const validEvents = [TOKEN_ENTER, TOKEN_EXIT, TOKEN_MOVE_IN, TOKEN_MOVE_OUT, TOKEN_MOVE_WITHIN,
```
aaclayton: Suggest defining this as a `static` just for cleanliness.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,68 @@
...
+  /** @override */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    const {TOKEN_ENTER, TOKEN_EXIT, TOKEN_MOVE_IN, TOKEN_MOVE_OUT, TOKEN_MOVE_WITHIN,
+      TOKEN_TURN_START, TOKEN_TURN_END, TOKEN_ROUND_START, TOKEN_ROUND_END} = CONST.REGION_EVENTS;
+    const validEvents = [TOKEN_ENTER, TOKEN_EXIT, TOKEN_MOVE_IN, TOKEN_MOVE_OUT, TOKEN_MOVE_WITHIN,
+      TOKEN_TURN_START, TOKEN_TURN_END, TOKEN_ROUND_START, TOKEN_ROUND_END];
+    const {id, name, img, description, effects, tags} = crucible.api.models.CrucibleAction.defineSchema();
+    return {
+      actionIdentifier: new fields.StringField({initial: null, required: true, nullable: true}),
```
aaclayton: Is the idea here that this is a "responsible action" or a "causing action"? There's a bunch of concepts that we maybe need to glue together somehow:

1. The `Actor` who performed the action
2. Which `CrucibleAction` they performed
3. The specific `ChatMessage` of the *instance* of the action performed
4. An `ActiveEffect` responsible for managing the region's persistence
5. The `Region` created by the action

It seems that we are linking this region behavior to #2, but maybe we should be linking it to #3 or perhaps even better #4. If we need to answer the question "what action did this region come from", the active effect that governs its persistence might be the right bridge to go over.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,68 @@
...
+      actionToPerform: new fields.SchemaField({
+        id, name, img, description, effects, tags
+      }, {required: true, initial: {id: "action", name: "Action", img: "icons/svg/hazard.svg", effects: [], tags: []}}),
+      actor: new fields.DocumentUUIDField({type: "Actor"}),
+      affectedActors: new fields.TypedObjectField(new fields.NumberField({integer: true, nullable: false}), {
```
aaclayton, line 35: I don't have a clear vision of how/why we would use this. The idea is that the persistent AOE would forever only apply to a specific set of actors? I'm struggling to think of why we would want that.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,68 @@
...
+        expandKeys: false,
+        validateKey: uuid => {
+          const {id, type} = foundry.utils.parseUuid(uuid);
+          if ( (type !== "Actor") || !foundry.data.validators.isValidId(id) ) return false;
+        }
+      }),
+      events: this._createEventsField({events: validEvents, initial: ["tokenEnter", "tokenTurnStart"]}),
+      oncePerRound: new fields.BooleanField({initial: true, required: true, nullable: false})
```
aaclayton: Some ambiguity around whether this means once-per-round globally or once-per-round-per-token. I assume the later, but it invites some question.

module/models/spell-action.mjs
```diff
@@ -135,6 +135,28 @@ export default class CrucibleSpellAction extends CrucibleAction {
         this.img = this.rune.img;
         this.description = _loc("ACTION.DEFAULT_ACTIONS.Cast.Description"); // TODO make dynamic
       }
+      if ( this.hasPersistentRegion ) {
```
aaclayton: It seems to me that we should get this at a more general level on the parent class without needing this code in the `CrucibleSpellAction` specifically.

module/models/action.mjs
```diff
@@ -646,22 +664,27 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
         limit: new fields.NumberField({required: false, nullable: false, initial: undefined, integer: true, min: 1}),
         self: new fields.BooleanField()
       }),
+      regionBehavior: new fields.SchemaField({
+        name: new fields.StringField(),
+        system: new fields.SchemaField({
```
aaclayton, line 669: We are implicitly associating the region behavior as a specific type of region beheavior (rather than this being able to be any type of behavior), so maybe it's not so useful to have the extra level of `system` data encapsulation?

module/models/action.mjs
```diff
@@ -826,6 +849,15 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
     return SYSTEM.ACTION.TARGET_TYPES[this.target.type]?.region && !this.region;
   }
 
+  /**
+   * Is this action configured to place a non-ephemeral region?
+   * @type {boolean}
+   */
+  get hasPersistentRegion() {
```
aaclayton, line 856: We may want to require that the action bears a self-target effect on the actor. In order for a region to be persistent we need to require that it bears an ActiveEffect for its duration.

module/models/action.mjs
```diff
@@ -1408,8 +1440,31 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
     // Create the RegionDocument, visible to GMs and the placing user
     if ( this.region ) {
       const regionData = this.region.toObject();
+      regionData.name = this.name;
       regionData.ownership = {default: 0, [game.user.id]: CONST.DOCUMENT_OWNERSHIP_LEVELS.OBSERVER};
       regionData.visibility = CONST.REGION_VISIBILITY.OBSERVER; // Author and GM only until confirmed
+      if ( this.hasPersistentRegion ) {
```
aaclayton, line 1446: If we are creating a persistent region, we need to make sure that the created region ID gets registered to the ActiveEffect that models its duration.

aaclayton (changes requested):
Good stuff @roth-michael, reviewing this once again has provoked a number of questions though. Some tough design calls to make here. 

This is starting to get very big, so we should also consider whether there are pieces of this that can be carved off and merged without controversy.

module/models/behavior-action.mjs
roth-michael: The idea here is that it gives access to the action responsible for the region's creation. Primary reason for this is to be able to infer who to target based on the targeting scope of the original action, if present. This needs to be separate from "what actor should be performing this" to allow for instances of a non-action-originating instance of this behavior (like setting up a trap manually in a region), since a performing actor is _always_ required but an originating action is not. I originally considered pointing to a chat message but thought that an action identifier was more trustworthy. Happy to iterate on it though.

module/models/behavior-action.mjs
roth-michael: Poorly named. This is for tracking the most recent round that each affected actor was affected, necessary for "once per round" functionality. It begins empty and is populated whenever the action is used.

crucible.mjs
roth-michael: I’ll give it some thought. It's effectively a "perform action" behavior I suppose.

module/models/action.mjs
roth-michael: On this note: what about something like placing a trap? In theory this could utilize the framework in this PR, but wouldn't have a duration and therefore _might_ not necessitate an effect?

module/models/action.mjs
roth-michael: This should be automatic elsewhere, we're getting this behavior already.
Specifically it happens in `confirm`.

module/models/spell-action.mjs
roth-michael: I’ll review why I thought this needed to be spell-specific tomorrow. Agree it would be better to have it be more general.

module/models/behavior-action.mjs
roth-michael: In the actual configuration this would I think have a tooltip. For this (and perhaps the to-be-renamed `affectedActors`) do you think a comment would be sufficient? Struggling to think of a better name than `oncePerRound` that's doesn't strain acceptable limits of verbosity.

module/models/spell-action.mjs
roth-michael: Taking a closer look, I think there are two elements:
1. `id` needs to be split by `.` and camel cased accordingly (the action identifier cannot start `spell.something` or it will be identified as a composed spell during usage). This I can generalize to the existing logic in `action.mjs`.
2. Setting tags. The reason this isn't a factor for non-spells is that any non-spell aura that needs a region-bound action will have that action configured on the originating action. This isn't possible with composed `aura` spells, since there is no designated action (just the Cast Spell action). This can probably be pulled out into the `aura` spellcraft hook for cleanliness (and later the `wall` hook if needed).

module/models/action.mjs
roth-michael: Re-examining this: Part of the benefit of doing it this way is that one can then directly treat `someAction.regionBehavior` as partial creation data for a region behavior. If we didn't nest the `system` fields under `system` in the schema itself, we'd need to do so prior to doing any data operations later (for instance when opening the behavior config sheet from the Action config, or when merging pre-configured info onto a default during action use). Not sure if it's a good change to remove the one layer of nesting at the expense of losing that benefit.

module/models/action.mjs
roth-michael: Plus, I just remembered, in `#recordEffectEvents` we _create_ a self-targeting effect if one doesn't exist and a persistent region is requested. So we actually approach this from the opposite direction.

roth-michael:
@aaclayton latest commit addresses comments I felt good about addressing, namely:
- Changed `crucible.persistentAOE` to `crucible.action`, referring to it as "Crucible Action" or "Crucible Action Behavior" (the latter in places where it would cause confusion with the concept of a normal Action)
- Threw in comments to clarify `affectedActors` and `oncePerRound` in the schema definition for the behavior
- Removed aura-gesture-specific logic from `CrucibleSpellAction`:
  - Now, if the action should create a persistent region, inside `_prepare` a default `regionBehavior` is set. (I would have preferred this in `_prepareData`, but `hasPersistentRegion` will be `null` for a composed spell at that point and I felt more comfortable moving this small logic into `_prepare` than I did moving the `super._prepareData()` call around)
  - Inside the `prepare` hook for the `aura` component, tags are set on that default
  - Slight adjustment to our `generateId` such that `.` symbols are treated the same as spaces, so `generateId("spell.life.auraRegion")` becomes `"spellLifeAuraRegion"`, rather than `"spelllifeauraRegion"`. I don't think this is _necessary_ but I do think it is prettier, and it shouldn't break anything.
- Made a private static `#VALID_EVENTS` to clean up the behavior schema creation (and did the same for the Hazard behavior, which uses the same set of events)

Things I did _not_ address, and would like further feedback on/response to my responses before making further changes:
- I've kept the link directly to the `actionIdentifier`. I think I'd be happy to link it to the Active Effect instead, but it again may lead to some extra legwork. We only care about what the targeting of the original action was, and in the case of a composed spell directly use the identifier to re-create the spell action, so unless it's an ideological "this is the correct way to traverse back to said action" - which it may be - I feel like keeping the identifier is more efficient for us. See the only spot we use it:
```js
// Skip invalid targets
const actionContext = {actor: sourceActor, region: this.parent.parent};
if ( this.actionIdentifier?.length ) {
  const originAction = this.actionIdentifier.startsWith("spell.")
    ? crucible.api.models.CrucibleSpellAction.fromId(this.actionIdentifier, actionContext)
    : sourceActor.actions[this.actionIdentifier]?.clone({}, actionContext);
  const validTargets = new Set(originAction?.acquireTargets().keys() ?? []);
  if ( !validTargets.has(actor) ) return;
}
```
- I did not rename `affectedActors` or `oncePerRound` - I think the commenting is sufficient for both from a code perspective, and only the latter is surfaced via the UI (and when it is, it has a tooltip). Happy to rename them if you have better names in mind, though; always prefer self-documenting field names.
- I did not de-nest the `system` field in the `regionBehavior` `SchemaField` on an Action. See my reasoning [here](https://github.com/foundryvtt/crucible/pull/1008#discussion_r3447384752). Again, can change if you think it _should_ change, but I think it'd just add some extra destructuring & logic in a couple of spots with little gain.
- I did not change the `hasPersistentRegion` getter to mandate a self-target effect; in `#recordEffectEvents` if `hasPersistentRegion` is true and there is no self-target effect event, we _create one_. So whether it should be a persistent region can be derived entirely without consideration of effects.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    const {id, name, img, description, effects, tags} = crucible.api.models.CrucibleAction.defineSchema();
+    return {
+      actionIdentifier: new fields.StringField({initial: null, required: true, nullable: true}),
+      actionToPerform: new fields.SchemaField({
```
aaclayton: I think there's still a little ambiguity in `actionIdentifier` in that - without looking up the localization label/hint it's unclear whether this is the identifier of the origin action or the identifier of the action to perform. 

My suggestion would be to simplify naming to `origin`, `action`. `origin` describes the origination of the region. `action` describes the mechanics the region causes.

More terse and less ambiguous imo.

FURTHERMORE. I'm still not sold on the data type of `actionIdentifier`. I think we will need a way to glue together the *instance of the action* to the region it created, for example consider some sort of action where the behavior of its region depends on whether the original action was a critical success or not. I think we need to know more than just which prototype action was used, we perhaps need to track some details of the outcome of that action. 

I really think the ActiveEffect is the right glue here. An originated action region should have an AE behind it, so I think the region behavior `origin` should be a UUIDField that points to an `ActiveEffect` instance. It would be the responsibility of the ActiveEffect to contain or point towards any further data about what the action was.

crucible.mjs
```diff
@@ -896,7 +906,7 @@ async function packageCompendium(documentName, packName, folder) {
  * @returns {string}          A standardized camel-case ID
  */
 function generateId(title, length) {
-  const id = title.split(" ").map((w, i) => {
+  const id = title.split(" ").flatMap(w => w.split(".")).map((w, i) => {
```
aaclayton: This specific change should really be a separate PR.

system.json
```diff
@@ -262,7 +262,8 @@
     "RegionBehavior": {
       "crucible.hazard": {
         "htmlFields": ["description"]
-      }
+      },
+      "crucible.action": {}
```
aaclayton: Does it need `htmlFields` declared?

aaclayton (changes requested):

module/models/behavior-action.mjs
roth-michael: Makes sense - I like the name changes, and I’ll try routing through the AE to get the originating action. Though my initial thinking is we needn't care about the results of the specific action ever; if something should modify the region action, we should handle that in the action's hooks. But still, AE solution is good, and ensures we set up a framework for being able to retrieve the origin action of any effect (assuming we don't already have that - need to check)

crucible.mjs
roth-michael: Yeah, alright. Doesn't really matter if it's `spelllifeauraRegion` or `spellLifeAuraRegion` right now, I’ll remove this and open it separately.

system.json
roth-michael: Yes - good shout, thanks.

roth-michael:
Alright, next round of feedback addressed. Added `htmlFields` and removed the out-of-scope `generateId` change, now onto the meat: `actionToPerform` -> `action`, and `actionIdentifier` -> `origin`, the latter of which has a laundry list of relevant changes:
- It is now a `DocumentUUIDField` for the `ActiveEffect` which is tracking the existence of the region (if any). 
- This field will always be populated for an Aura (or any other action-spawned behavior), but will not ever be modifiable via the UI. It will be shown for non-synthetic regions (hidden when configuring the behavior directly on an action)
- Since we're no longer just passing a string, each time we relied on having the identifier we instead rely on getting the action itself (more on that after this sub-bullet)
  - When modifying a synthetic behavior (on-action config) we must pass an ephemeral `actionId` flag in addition to `itemUuid` so that the proper action gets hit with an item update

Right now, I've set up a very basic `CrucibleBaseActiveEffect#getOriginAction` and supporting logic. Long-term, I suspect we'll want to do something like what we have with being able to rebuild an action from a chat message & flags, but on an effect. _For now_, I've implemented the following (and will continue bullet points with downstream effects of the new implementation):
- In `CrucibleActor#_applyActionEffects`, allow an `originAction` option which is a `CrucibleAction`. If passed, set `flags.crucible.originAction` to that action's `id` on every effect being created/updated
- In `CrucibleAction##applyEvents`, pass `originAction: this` to the above function
- Implement `CrucibleBaseActiveEffect#getOriginAction` which takes an optional `CrucibleActionContext`, then tries to grab the action with the flagged ID on the parent effect off the _origin actor_ of the effect (based on the parent effect's `origin` field), cloned with that action context if provided (in the case of Spell actions, using `fromId` instead of grabbing off the source actor). I've made this sync and exit early if the actor is not an Actor document with subtype extending `CrucibleBaseActor`. Not sure if maybe it should be async. Eventually I think this would be where we compiled more comprehensively stored action data & create an action from it
- Inside `CrucibleActionRegionBehavior#_handleRegionEvent`, we check for the existence of the origin effect and, if present, get the origin action to use for acquiring targets
- I split out the "use some self event to record the existence of a persistent region" logic from the "update region visibility & behavior disabled status (& now, setting the origin uuid) so that the first bit happens _prior_ to effect creation, and the latter bit _post_ effect creation. Reason being: Events will trigger potentially immediately upon enabling the region behavior. Therefore the origin effect _must exist at that point_ to be referenced. As a consequence of this, the previous logic (which updated region visibility depending on reversal) has been amended such thot, during reversal, _nothing_ happens to the region. This is because a persistent region will, by this point, be slated for deletion already as a downstream effect of deleting the tracking active effect, so odds are high (if not certain?) that trying to call `update` on this doomed region will throw an error.
  - Part of this change involved moving the `retained` check up prior to the "just set the first self effect to track it" logic, which means a hook could theoretically specify a _specific_ self-effect (in the event that there are multiple) to be the region-tracking one. Also means down the line we could potentially have an action that creates an aura on _somebody else_ and has the effect which tracks it on them too. Dunno if we ever will, but we could with this logic, I think.

Phew. Almost as much effort writing that up as making the changes themselves. Happy to go into more detail on any point of the above and continue iterating.

roth-michael:
Closed in favor of #1471, #1472, #1474, #1475

PR #1013, opened 2026-05-11
[#999] Display physical item properties as tags in inventory, render tooltips

### LINKED ISSUE #999: Include the full set of Item Properties as header tags when viewing a physical item and ensure that the tag tooltips are discoverable to learn their behaviors.
Debatably not a "bug" but just some bad UX.
Original report: https://discord.com/channels/170995199584108546/1116696333894570094/1502071209645441235

Ambush is in fact in the system rules _but_ "Intuitive" is _not_, so it should be added. And either we should add the properties as tags or, if that'd take up too much space, we should add tooltips on the checkboxes on the sheet.

roth-michael:
Closes #999

- All physical items will now render all their properties as tags in inventory/item header. "Investment" and "Stackable" are exceptions.
- Added "Intuitive" to the equipment rules journal, using the existing tooltip text for it.
- Ensured property tooltips were being prelocalized.
- Added property tooltips for Armor: Bulky (using rules text), Natural (made it up), Noisy (made it up). Needed to add Natural, otherwise it'd try to intelligently render a tooltip and end up using the _action_ tooltip. And at that point felt silly not to add Noisy.
- Gave consumables their own Thrown localization, since the weapon version's tooltip made no sense for a consumable.
- Fixed an issue where a second (or worse, third) row of tags on an item could mess up the header. Before & After:

image
image

### DIFF
--- a/module/models/item-physical.mjs
+++ b/module/models/item-physical.mjs
@@ -353,6 +353,12 @@ export default class CruciblePhysicalItem extends foundry.abstract.TypeDataModel
     if ( this.dropped ) tags.dropped = this.schema.fields.dropped.label;
     if ( this.requiresInvestment ) tags.invested = this.invested ? this.schema.fields.invested.label
       : _loc("ITEM.PROPERTIES.NotInvested");
+
+    // Properties
+    for ( const prop of this.properties ) {
+      if ( ["investment", "stackable"].includes(prop) ) continue;
+      tags[prop] = this.constructor.ITEM_PROPERTIES[prop];
+    }
     return tags;
   }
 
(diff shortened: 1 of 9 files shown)


module/models/item-physical.mjs
aaclayton, line 349: Does this property-based tag get duplicated by the properties loop?

aaclayton (approved):

aaclayton (approved):

module/models/item-physical.mjs
roth-michael: No, it does get clobbered though. If anything, could probably remove this line.

PR #1015, opened 2026-05-11
[#1006] Add hazard region behavior

### LINKED ISSUE #1006: Create a Crucible Hazard region behavior that implements the effects of a Hazard via a Scene Region

roth-michael:
Closes #1006 

image

I limited selectable tags to only the modifier tags, since everything else doesn't really make sense when we've got dropdowns for defense type, damage type, and resource. Wasn't sure whether we'd want to show the GM a roll dialog to let them do ad-hoc boons/banes, so decided an option for that would be best.

### DIFF
--- a/module/models/behavior-hazard.mjs
+++ b/module/models/behavior-hazard.mjs
@@ -1,0 +1,60 @@
+export default class CrucibleHazardRegionBehavior extends foundry.data.regionBehaviors.RegionBehaviorType {
+
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["REGION_BEHAVIORS.HAZARD"];
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    const {TOKEN_ENTER, TOKEN_EXIT, TOKEN_MOVE_IN, TOKEN_MOVE_OUT, TOKEN_MOVE_WITHIN,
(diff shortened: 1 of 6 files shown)


module/models/behavior-hazard.mjs
```diff
@@ -0,0 +1,60 @@
...
+    const {action, focus, heroism, ...resources} = foundry.utils.deepClone(SYSTEM.RESOURCES);
+    const tags = Object.values(SYSTEM.ACTION.TAGS).reduce((acc, t) => {
+      if ( t.internal || (t.category !== "modifiers") ) return acc;
+      const cat = SYSTEM.ACTION.TAG_CATEGORIES[t.category];
+      const group = cat?.label;
+      return {...acc, [t.tag]: {label: t.label, group}};
+    }, {});
+    return {
+      name: new fields.StringField(),
+      danger: new fields.NumberField({initial: 0, required: true, nullable: false, integer: true}),
```
aaclayton, line 26: I think we should add `description` as an `HTMLField` (with declared sanitization in the `system.json` manifest) so that hazards can have a short narrative description of what happens.

This could be reserved for a later PR, but it's probably a pretty quick add. That said, I don't know how well the auto-generated behavior configuration sheet handles an `HTMLField`

aaclayton (changes requested):

module/models/behavior-hazard.mjs
roth-michael: Handles it well:
image
Adding

aaclayton (approved):

PR #1021, opened 2026-05-12
[#1019, #1020] Re-render favorites on action CRUD, fix various counterspell bugs, add automation for Horrific Critical

### LINKED ISSUE #1019: Actor sheet sidebars (favorites) should re-render when an action is used, confirmed, reversed, or its chat message is deleted

### LINKED ISSUE #1020: Counterspell doesn't work out of combat, and doesn't conditionally show in favorites based on last action

roth-michael:
Closes #1019 
Closes #1020 
New stuff:
- On message create & delete if `flags.crucible.action` exists, and on update if `flags.crucible.confirmed` is being changed, re-render all actor sheet sidebars, ensuring accurate favorites
- Horrific Critical usage is blocked if your previous action was not a critically successful melee strike
- Combat-based counterspell is now blocked not only if it's your turn or you're out of combat, but also if there is no possible valid target. (In the future, we could make it even stricter; right now it'll let you use it on someone who's confirmed the action, though it'll yell at you if you try to confirm the counterspell afterwards)

Fixed stuff:
- Counterspell never got updated to use the new `targets` Map instead of array. Now it has. It also wasn't setting `this.targets` previously, now it is.
- `fontOfLife` hook moved before `gemOfConjuredFlame` because alphabets are cool.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -439,7 +439,7 @@
       "InvalidTarget": "You must designate at least one {type} target for {action}!",
       "InvalidTargetsGeneric": "Invalid targets for this action!",
       "LastCriticallyMissed": "You may only perform {action} after a basic Strike which did not critically miss.",
-      "LastNotMeleeCrit": "You may only perform {action} immediately after a Strike which critically succeeded.",
+      "LastNotMeleeCrit": "You may only perform {action} immediately after a melee Strike which critically succeeded.",
       "MaximumRange": "A selected target is further than the maximum range of {max} feet.",
       "MinimumRange": "A selected target is closer than the minimum range of {min} feet.",
       "MinimumSize": "You must be at least size {size} to use {action}",
--- a/lang/en.json
+++ b/lang/en.json
@@ -439,6 +439,7 @@
       "InvalidTarget": "You must designate at least one {type} target for {action}!",
       "InvalidTargetsGeneric": "Invalid targets for this action!",
       "LastCriticallyMissed": "You may only perform {action} after a basic Strike which did not critically miss.",
+      "LastNotMeleeCrit": "You may only perform {action} immediately after a Strike which critically succeeded.",
       "MaximumRange": "A selected target is further than the maximum range of {max} feet.",
       "MinimumRange": "A selected target is closer than the minimum range of {min} feet.",
       "MinimumSize": "You must be at least size {size} to use {action}",
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -14,6 +14,7 @@ export default class CrucibleChatMessage extends ChatMessage {
   /** @inheritDoc */
   _onCreate(data, options, userId) {
     super._onCreate(data, options, userId);
+    if ( foundry.utils.hasProperty(data, "flags.crucible.action") ) CrucibleChatMessage.#renderAllSheetSidebars();
     this.#autoConfirmMessage();
   }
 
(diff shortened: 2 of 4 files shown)


lang/en.json
```diff
@@ -439,6 +439,7 @@
       "InvalidTarget": "You must designate at least one {type} target for {action}!",
       "InvalidTargetsGeneric": "Invalid targets for this action!",
       "LastCriticallyMissed": "You may only perform {action} after a basic Strike which did not critically miss.",
+      "LastNotMeleeCrit": "You may only perform {action} immediately after a Strike which critically succeeded.",
```
aaclayton:
```suggestion
      "LastNotMeleeCrit": "You may only perform {action} immediately after a melee Strike which critically succeeded.",
```

module/documents/chat-message.mjs
```diff
@@ -23,11 +24,36 @@ export default class CrucibleChatMessage extends ChatMessage {
...
+    if ( foundry.utils.hasProperty(this, "flags.crucible.action") ) CrucibleChatMessage.#renderAllSheetSidebars();
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Re-render all currently-rendered actor sheet sidebars
+   * @returns {Promise<void>}
+   */
+  // TODO: Find a better place for this to live
```
aaclayton: I think you can remove the TODO, it's fine here for now.

module/documents/chat-message.mjs
```diff
@@ -23,11 +24,36 @@ export default class CrucibleChatMessage extends ChatMessage {
...
+  static #renderAllSheetSidebars() {
+    const promises = [];
+    for ( const app of foundry.applications.instances.values() ) {
+      if ( !(app instanceof crucible.api.applications.CrucibleBaseActorSheet) ) continue;
+      promises.push(app.render({parts: ["sidebar"]}));
```
aaclayton: You can use the handy `CrucibleBaseActorSheet.instances` generator!

module/hooks/action.mjs
```diff
@@ -476,6 +483,17 @@ HOOKS.healingTonic = {
 
 /* -------------------------------------------- */
 
+HOOKS.horrificCritical = {
+  canUse() {
+    const lastAction = this.actor.lastConfirmedAction;
+    if ( !lastAction?.tags.has("melee") || !lastAction?.events.some(e => (e.type === "strike") && e.isCriticalSuccess) ) {
+      throw new Error(_loc("ACTION.WARNINGS.LastNotMeleeCrit", {action: this.name}));
```
aaclayton, line 492: We have a specialized thrown warning string for horrificCritical, but counterspell is just `return false`. I'm not sure how scalable it is to have a custom canUse warning string for every action we ever do, but if we're doing custom canUse strings, then Counterspell probably deserves one.

module/hooks/action.mjs
```diff
@@ -247,8 +247,15 @@ HOOKS.counterspell = {
...
   },
+  canUse() {
+    if ( this.tags.has("noncombat") ) return;
+    const lastAction = ChatMessage.implementation.getLastAction();
+    const wasSpell = lastAction && (lastAction.tags.has("composed") || lastAction.tags.has("iconicSpell"));
+    if ( !wasSpell ) return false;
+  },
   async roll(target) {
     if ( this.usage.targetAction.message ) return;
+    this.events.findSplice(e => e.type === "spell");
```
aaclayton, line 260: Since this is destructive to the `events` stream, I think a single-line comment explaining why we do this would be helpful for readability

aaclayton (changes requested):
Just a couple small comments, good fixes!

aaclayton (approved):

PR #1022, opened 2026-05-12
[#1002] Add sense creature detection mode

roth-michael:
#1002 
Currently, the main intended use is with the Sense gesture. In theory, one could be granted this detection mode _without_ that by using `flags.crucible.senseRune` directly on an actor. No support currently for sensing using multiple runes simultaneously, though that's something we could add.

### DIFF
--- a/module/canvas/detection-modes/sense-creature.mjs
+++ b/module/canvas/detection-modes/sense-creature.mjs
@@ -1,0 +1,90 @@
+export default class DetectionModeSenseCreature extends foundry.canvas.perception.DetectionMode {
+
+  /**
+   * Glow colors for each rune.
+   * @type {Readonly<Record<string, [number, number, number, number]>>}
+   */
+  static #GLOW_COLORS = Object.freeze({
+    control: [0.588, 0.278, 1.0, 1],
+    death: [0.663, 1.0, 0.506, 1],
+    earth: [0.431, 0.796, 0.0, 1],
+    flame: [1.0, 0.424, 0.0, 1],
+    frost: [0.0, 0.98, 1.0, 1],
+    illumination: [1.0, 0.976, 0.749, 1],
+    illusion: [0.886, 0.063, 0.212, 1],
+    kinesis: [0.843, 0.843, 0.843, 1],
+    life: [1.0, 0.443, 0.565, 1],
+    lightning: [1.0, 0.725, 0.424, 1],
+    oblivion: [0.478, 0.043, 0.086, 1],
+    soul: [0.008, 0.608, 1.0, 1]
+  });
+
+  /**
+   * Cached glow filters keyed by rune id.
+   * @type {Map<string, foundry.canvas.rendering.filters.GlowOverlayFilter>}
+   */
+  static #filtersByRune = new Map();
+
+  /**
+   * The rune id of the most recently detected target. Set by `_canDetect` and consumed by `getDetectionFilter`,
+   * which the visibility loop calls immediately after a successful `testVisibility`.
+   * @type {string|null}
+   */
+  static #pendingRune = null;
+
+  /**
+   * Which creature categories can be seen by which rune id.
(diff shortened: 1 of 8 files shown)


module/canvas/detection-modes/sense-creature.mjs
```diff
@@ -0,0 +1,90 @@
...
+    earth: [0.431, 0.796, 0.0, 1],
+    flame: [1.0, 0.424, 0.0, 1],
+    frost: [0.0, 0.98, 1.0, 1],
+    illumination: [1.0, 0.976, 0.749, 1],
+    illusion: [0.886, 0.063, 0.212, 1],
+    kinesis: [0.843, 0.843, 0.843, 1],
+    life: [1.0, 0.443, 0.565, 1],
+    lightning: [1.0, 0.725, 0.424, 1],
+    oblivion: [0.478, 0.043, 0.086, 1],
+    soul: [0.008, 0.608, 1.0, 1]
```
aaclayton: It might be more maintainable if we write this as something like:
```
control: Color.from("#hex123").rgba
```
That would make it easier to preview and change the color in an IDE.

module/canvas/detection-modes/sense-creature.mjs
```diff
@@ -0,0 +1,90 @@
...
+   * @type {string|null}
+   */
+  static #pendingRune = null;
+
+  /**
+   * Which creature categories can be seen by which rune id.
+   * TODO: Determine whether this should live elsewhere. Maybe a reverse association from creature category to rune
+   * @type {Record<string, string[]>}
+   */
+  static #creaturesByRune = {
```
aaclayton: I'd like you to add a `sense {string}` key to every `CREATURE_CATEGORIES` to reflect "which rune can sense this creature type"

module/canvas/detection-modes/sense-creature.mjs
```diff
@@ -0,0 +1,90 @@
...
+  /* -------------------------------------------- */
+
+  /** @override */
+  _canDetect(visionSource, target) {
+
+    // Sense only works on creatures (for our purposes)
+    if ( !(target instanceof foundry.canvas.placeables.Token) ) return false;
+
+    // Check which rune is in use
+    // TODO: Determine the ideal way of storing this info
```
aaclayton: I think storing it on the sense effect makes good sense. I think you can remove the TODO, this is ideal (enough)

module/canvas/detection-modes/sense-creature.mjs
```diff
@@ -0,0 +1,90 @@
...
+    const source = visionSource.object.document;
+    const senseEffectId = SYSTEM.EFFECTS.getEffectId("sense");
+    const senseEffect = source.actor?.effects.get(senseEffectId);
+    const runeId = senseEffect?.flags.crucible?.senseRune ?? source.actor?.flags.crucible?.senseRune;
```
aaclayton: Maybe more generally just `flags.crucible.rune`? Or possibly we should plan ahead on a `runes` array which might allow multiple runes be used together for some sort of iconic spell "omni-sense"?

module/hooks/spellcraft.mjs
```diff
@@ -149,12 +150,31 @@ HOOKS.sense = {
...
+            enabled: true,
+            range: this.target.size
+          }
+        }]
+      },
+      flags: {
+        crucible: {
+          senseRune: this.rune.id
+        }
+      }
```
aaclayton: Maybe instead of using AE changes we could instead recognize that "senseCreature" is a core system sense type, and we could have a standardized `CrucibleToken` data preparation workflow that checks for an active sense?

Using hooks is obviously the most modular and extensible, but this is a case (like certain item properties or effect statuses) where I think we can directly encode the logic in the system. There could be other talents that also give the ability to sense certain types of creatures, not exclusively using the Gesture: Sense spell.

aaclayton (changes requested):

aaclayton (approved):

PR #1024, opened 2026-05-12
make default targets for checks automatically extract actors from groups

### LINKED ISSUE #1023: Improve the ergonomics of skill checks and hazards when a party token is controlled. When a party token is selected, it is used as the target for checks instead of the tokens in it
This was pretty annoying on the region map in Ember - when doing a check, instead of the members of the party, the default actor for the check was the party group actor itself.

danielrab:
closes https://github.com/foundryvtt/crucible/issues/1023


aaclayton:
Resolved via https://github.com/foundryvtt/crucible/commit/f9b05d40e0e9f5c9c6c4e16a7ca8466bd3d06f99

PR #1027, opened 2026-05-13
Fix actor hooks never running

roth-michael:
Opened this as a PR instead of directly committing because we may prefer a different solution to this: keeping the existing code, and adding `actorHooks` as a non-persisted field on the schema of the hero & adversary models. I don't know that there's much benefit to doing it that way, but still.


aaclayton (approved):
This is okay for now, I think we might do better with some sort of static metadata or non-persisted schema field in the future, but this is safe.

PR #1036, opened 2026-05-13
[#1031] Represent unknown/unset DCs as `null` instead of `0`, and allow group check DCs to be modified after message post

### LINKED ISSUE #1031: Add an intentional difference between DC=0 and DC=undefined
This seems to be a purely visual issue, the result in the attached image is still considered a crit for automation, but it's rendered as if it's a glancing blow in terms of the colors used.
Image

roth-michael:
Closes #1031 
0DC skill checks mean "show as unknown DC, and allow the GM to set via the chat message's context menu." 0DC attack rolls mean "the poor fella getting smacked somehow actually has 0 defense," and therefore should render as such.

### DIFF
--- a/module/chat.mjs
+++ b/module/chat.mjs
@@ -23,30 +23,35 @@ export const TEMPLATES = {
 export function addChatMessageContextOptions(_app, options) {
   if ( !game.user.isGM ) return;
 
-  // Assign difficulty for skill checks
+  // Assign difficulty for skill or group checks
   options.push({
     label: _loc("DICE.SetDifficulty"),
     icon: '<i class="fas fa-bullseye"></i>',
     visible: li => {
       const message = game.messages.get(li.dataset.messageId);
       const flags = message.flags.crucible || {};
-      return message.isRoll && flags.skill;
+      return message.isRoll && (flags.skill || flags.groupCheck);
     },
     onClick: async (_e, li) => {
       const message = game.messages.get(li.dataset.messageId);
-      const roll = message.rolls[0];
+      const flags = message.flags.crucible;
+      const skills = new Set(message.rolls.map(r => r.data.type));
       const formData = await foundry.applications.api.DialogV2.input({
         window: {title: _loc("DICE.SetDifficulty")},
-        content: `\
+        content: Array.from(skills).map(skill => `\
         <div class="form-group slim">
-            <label>DC Target</label>
-            <div class="form-fields">
-                <input type="number" name="dc" value="${roll.data.dc || 15}" autofocus>
-            </div>
-        </div>`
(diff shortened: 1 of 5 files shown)


aaclayton (changes requested):
Not sure I'm liking this solution. I think probably any `Number.isNumeric(dc)` should count as a set DC and we need some other convention to express an unknown DC in the skill check case, probably explicit `null` would be best.

Want to take a crack at shifting this so that numeric DCs are always treated as "set" and we intentionally set `null` when we want the DC to be initially undetermined?

roth-michael:
Updated description of changes:
- `0` DC now means `0` DC. Unset DC will be `null`
- Aggregate group checks now display `???` instead of score if any of the contained checks have a `null` DC
- Changed `GroupCheck##renderGroupCheckCard` to `GroupCheck.renderGroupCheckCard`. It didn't need to be an instance method since it didn't use `this`, and it needed to be un-privated so that updates to a group check card's DCs could properly re-render the content
- As alluded to above, can now change the DC of each skill in a group check via the context menu, just as you can for single skill rolls

module/dice/standard-check.mjs
```diff
@@ -315,7 +315,7 @@ export default class StandardCheck extends Roll {
     if ( targetLabel === undefined ) {
       const skill = SYSTEM.SKILLS[this.data.type];
       const label = skill?.label ?? defenseType;
-      const dcLabel = game.user.isGM ? (dc || "??") : "";
+      const dcLabel = game.user.isGM ? (`${dc}` ?? "??") : "";
```
aaclayton, line 318: `dc` is now never nullish:

```suggestion
      const dcLabel = game.user.isGM ? (Number.isNumeric(dc) ? String(dc) : "??") : "";
```

module/chat.mjs
```diff
@@ -23,30 +23,35 @@ export const TEMPLATES = {
...
+        content: Array.from(skills).map(skill => `\
         <div class="form-group slim">
-            <label>DC Target</label>
-            <div class="form-fields">
-                <input type="number" name="dc" value="${roll.data.dc || 15}" autofocus>
-            </div>
-        </div>`
+          <label>${_loc("DICE.DCTargetSpecific", {skill: SYSTEM.SKILLS[skill]?.label ?? skill})}</label>
+          <div class="form-fields">
+            <input type="number" name="${skill}" value="${message.rolls.find(r => r.data.type === skill).data.dc ?? 0}">
```
aaclayton, line 45:
```suggestion
            <input type="number" name="${skill}" value="${message.rolls.find(r => r.data.type === skill).data.dc ?? 0}" autofocus>
```

aaclayton (approved):

PR #1058, opened 2026-05-15
Token plummet implementation

Fyorl:
1. Applies flying, hovering, or falling status effect at the conclusion of a movement-tagged action.
2. Adds a new 'fall' action type.
3. Chains a fall action onto the end of a confirmed movement-tagged action.
4. Rolls a hazard against the creature using the appropriate rules for falling after calculating distance to the surface.
5. Plummets the token on confirmation, dealing damage, knocking prone, clearing the falling status.


module/const/action.mjs
```diff
@@ -939,7 +940,7 @@ export const TAGS = {
     priority: 9,
     prepare() {
       this.usage.hasDice = true;
-      this.usage.bonuses.enchantment = this.item.system.config?.enchantment.bonus || 0;
+      this.usage.bonuses.enchantment = this.item?.system.config?.enchantment.bonus || 0;
```
Fyorl, line 943: Fixes an error when using the `generic` tag on an action that has no associated item.

module/documents/chat-message.mjs
```diff
@@ -25,7 +32,7 @@ export default class CrucibleChatMessage extends ChatMessage {
     super._onUpdate(data, options, userId);
     const flags = this.flags.crucible || {};
     if ( foundry.utils.hasProperty(data, "flags.crucible.confirmed") ) CrucibleChatMessage.#renderAllSheetSidebars();
-    if ( flags.action && flags.vfxConfig && (data.flags.crucible.confirmed === true) ) this.#playVFXEffect();
+    if ( flags.action && flags.vfxConfig && (foundry.utils.getProperty(data, "flags.crucible.confirmed") === true) ) this.#playVFXEffect();
```
Fyorl, line 35: Might not be necessary anymore, this was added before I found the root cause of my double-reverse issue.

module/documents/token.mjs
```diff
@@ -172,9 +172,61 @@ export default class CrucibleToken extends foundry.documents.TokenDocument {
   async #revertUndoneMovement(movementId) {
     const message = game.messages.contents.findLast(m => m.flags?.crucible?.movement === movementId);
     if ( !message ) return;
+    // The action's own reverse flow is already handling this message; do not re-reverse or delete it.
+    if ( message._reversing ) return;
```
Fyorl, line 176: If you reverse a chat message that has movement, you end up hitting this method, which also tries to reverse the same chat message:
`#applyEvents` -> `#applyMovement` -> `token.revertRecordedMovement` -> `_onUpdateMovement` -> `#revertUndoneMovement`

module/const/statuses.mjs
```diff
@@ -99,6 +99,20 @@ export const statusEffects = {
...
     page: "Compendium.crucible.rules.JournalEntry.crucibleConditio.JournalEntryPage.burrowing0000000"
   },
+  flying: {
+    id: "flying",
+    name: "ACTIVE_EFFECT.STATUSES.Flying",
+    img: "icons/svg/wing.svg",
+    hud: nonGroupTypes,
+    page: "" // TODO
+  },
+  hovering: {
```
aaclayton: Did we end up *needing* hovering, or can we get by only with flying for now?

module/const/action.mjs
```diff
@@ -1244,6 +1267,12 @@ export const TAGS = {
         const isNonCrawlMovement = this.movement?.waypoints?.some(w => w.action !== "crawl");
         if ( isNonCrawlMovement ) await this.actor.toggleStatusEffect("prone", {active: false});
       }
+    },
+    async postConfirm(reverse) {
```
aaclayton, line 1271: Want to check in with you to understand what necessitates a `postConfirm` rather than just a standard `confirm` behavior.

module/documents/chat-message.mjs
aaclayton: Let's try reverting this then if its not a strictly necessary change.  Probably we can just use optional chaining to be safe.

module/documents/token.mjs
```diff
@@ -172,9 +172,61 @@ export default class CrucibleToken extends foundry.documents.TokenDocument {
...
+    if ( !surfaces.length ) return null;
+    const points = this.getContainmentTestPoints(position);
+
+    // Walk surfaces from highest to lowest and return the first whose footprint contains 75%+ of the token.
+    // Scene#getSurfaces already orders surfaces by elevation.
+    for ( let i = surfaces.length; i--; ) {
+      const surface = surfaces[i];
+      if ( surface.elevation > elevation ) continue;
+      const inside = points.filter(p => surface.region.polygonTree.testPoint(p)).length;
+      if ( (inside / points.length) >= .75 ) return surface;
```
aaclayton: Probably not a huge performance concern, BUT I think we could optimize this. Instead of filtering and testing every test point, we would know ahead of time what is the critical threshold number of hits that would be a stable surface, similarly we would know the critical threshold number of misses that would rule out a surface from being stable.

Suppose there are 16 test points. If we miss on the first 5 tests, we don't need to test the remaining 11 because we know we have missed the surface. This could be a pretty big reduction in the number of tests for tokens that have a lot of points.

Alternatively, or additionally, is it worth doing a fast bounds overlap test before doing individual test point checks so we can quickly reject surfaces that don't overlap with the token hitbox?

module/documents/token.mjs
```diff
@@ -172,9 +172,61 @@ export default class CrucibleToken extends foundry.documents.TokenDocument {
...
+    }
+    return null;
```
aaclayton: If there is no defined surface beneath you, and if your `position.elevation` is greater than `canvas.level.elevation.base`, maybe we should treat the flat plane of `canvas.level.elevation.base` as a fallback "supporting surface"? 

The current functionality seems to require defining a full scene-wide surface rectangle for your base ground elevation, and while I imagine this is now considered **best practice** I'm not certain it should be mandatory. Thoughts?

module/hooks/action.mjs
```diff
@@ -373,6 +373,48 @@ HOOKS.electrochargeAmpoule = {
 
 /* -------------------------------------------- */
 
+HOOKS.fall = {
+  canUse() {
+    if ( !this.actor.statuses.has("falling") ) return false;
+  },
+  prepare() {
```
aaclayton, line 381: The `prepare` hook runs eagerly on every actor data preparation, which is probably more work than we need to do at this time. There's some things you could do here in prepare like configuring the damage type or certain usage bonuses, but I think it would be best to defer the work of `_findSupportingSurface` until `preActivate` which only happens lazily when the action is actually being used.

module/hooks/action.mjs
```diff
@@ -373,6 +373,48 @@ HOOKS.electrochargeAmpoule = {
...
+    Object.defineProperty(this, "movement", { value: movement, configurable: true });
+    this.recordEvent({ movement: movement.id, target: this.actor, type: "movement" });
+  },
+  configureVFX(vfxConfig) {
+    return crucible.api.canvas.vfx.landing.configureLandingVFXEffect(this) ?? vfxConfig;
+  },
+  finalizeVFX(vfxEffect, references) {
+    crucible.api.canvas.vfx.landing.finalizeLandingVFXEffect(this, vfxEffect, references);
+  },
+  postActivate() {
```
aaclayton: Slight code style preference to write hooks in their order of invocation, so placing configureVFX and finalizeVFX after postActivate.

module/models/action.mjs
```diff
@@ -523,6 +523,8 @@ class CrucibleActionTags extends Set {
  * - `CrucibleAction##applyEvents` - Walk the event stream and apply resource changes to each actor. When
  *   `reverse=true`, all deltas are inverted, effects are removed, and item snapshots are restored.
  * - `CrucibleAction##recordHeroism` - Award heroism for confirmed damage-dealing actions.
+ * - Action hooks called: `postConfirm`
+ *    - Fires after `#applyEvents` and the message-confirmed flag flip; used for chaining actions.
```
aaclayton, line 527: I suppose this has answered my prior question about what is the need for postConfirm being that you want the chat message confirmation flag to have already settled before triggering a next action - I think this makes sense to me.

module/models/action.mjs
```diff
@@ -2285,6 +2287,17 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
+    // Post-confirm hooks fire after #applyEvents has run and the message-confirmed flag is committed. Used for chaining
+    // actions.
+    for ( const test of this._tests() ) {
+      if ( typeof test.postConfirm !== "function" ) continue;
+      try {
+        await test.postConfirm.call(this, reverse);
+      } catch ( cause ) {
+        console.error(new Error(`"${this.id}" postConfirm failed`, { cause }));
+      }
+    }
```
aaclayton, line 2300: We could maybe have a helper function for this type of block that could be used multiple cases, something like `this.#callHooks("postConfirm", reverse);`

Might be a good ergonomics improvement, could be a follow-up PR, no need to change this now just a thought.

aaclayton (changes requested):
Looking nice @Fyorl some feedback!

module/const/statuses.mjs
Fyorl: Not needed. I only added it because I thought we wanted it hooked up as a special status.

module/const/action.mjs
Fyorl: `confirm` hooks fire before `#applyEvents`, so the actor doesn't have the falling status. So we either add a hook after `#applyEvents`, or we duplicate the work we already did in the movement hook to work out if the actor is falling.

module/documents/chat-message.mjs
Fyorl: Well, this is the equivalent of optional chaining

module/documents/token.mjs
Fyorl: We could try. `canvas.level` is the viewed level though, so it would be a different outcome depending on what level you're viewing. We could use the token's level maybe. Or we could do something like the lowest level in the scene. The latter seems like the most intuitive to me, but I wonder if we would have cases of tokens plummeting through solid ground into a dungeon level? Presumably in such a setup you would have a surface? But I'm not sure.

aaclayton: @Fyorl can we do the lowest level that is visible from your current level (which might be your current level - especially important for single-level scenes)? I think we can assume that if the user has gone to the trouble of setting up multiple mutually visible levels they have put surfaces where there need to be surfaces.

module/const/action.mjs
```diff
@@ -1384,6 +1413,17 @@ export const DEFAULT_ACTIONS = Object.freeze([
...
 
+  // Fall
+  {
+    id: "fall",
+    name: "ACTION.DEFAULT_ACTIONS.Fall.Name",
+    img: "systems/crucible/icons/statuses/falling.svg",
+    description: "ACTION.DEFAULT_ACTIONS.Fall.Description",
+    target: { type: "self", scope: 1 },
+    cost: { action: 0 },
+    tags: ["generic"]
```
aaclayton: I still think it would be worth adding the prone effect in the `effects` array here since it can be statically configured to only occur on a success. Save some logic later.

module/hooks/action.mjs
```diff
@@ -373,6 +373,47 @@ HOOKS.electrochargeAmpoule = {
...
 
+HOOKS.fall = {
+  canUse() {
+    if ( !this.actor.statuses.has("falling") ) return false;
+  },
+  prepare() {
+    this.usage.damageType = "bludgeoning";
+    this.usage.bonuses.base = 1;
+    // Apply prone unless the hazard is resisted.
+    this.effects = [{ statuses: ["prone"], result: { type: "success" } }];
```
aaclayton: This could be moved to the static action definition in `DEFAULT_ACTIONS`.

module/hooks/action.mjs
```diff
@@ -373,6 +373,47 @@ HOOKS.electrochargeAmpoule = {
 
 /* -------------------------------------------- */
 
+HOOKS.fall = {
+  canUse() {
+    if ( !this.actor.statuses.has("falling") ) return false;
+  },
```
aaclayton, line 380: We might want to add the `distance <= 0` condition as part of `canUse` so that you can't actually perform the fall action unless you have some distance that is necessary to fall?

module/hooks/action.mjs
```diff
@@ -373,6 +373,47 @@ HOOKS.electrochargeAmpoule = {
...
+    // Apply prone unless the hazard is resisted.
+    this.effects = [{ statuses: ["prone"], result: { type: "success" } }];
+  },
+  preActivate() {
+    const surface = this.token?._findSupportingSurface();
+    const distance = surface ? (this.token._source.elevation - surface.elevation) : 0;
+    this.usage.fallDistance = distance;
+    this.usage.bonuses.ability = distance;
+    this.usage.defenseType = distance > 30 ? "fortitude" : "reflex";
+    if ( distance > 30 ) this.tags.add("severe");
```
aaclayton: An important part of the falling rules is that falls of 10 feet or less deal no damage, so if the distance is <= 10 we could do `this.tags.delete("generic")` and `this.tags.add("harmless")` so that there is no dice roll for the fall and it only does the actual movement component with no damage roll.

aaclayton (approved):
Approved with some final comments before you merge it. Great work Kim!

module/hooks/action.mjs
Fyorl: You shouldn't have the falling status in that case then. It's possible to add manually I suppose, but I was intending that we can rely on this status as an indicator that the logic has already been run and made the determination, so that we don't have to repeat the logic in multiple places (and call `_findSupportingSurfaces` which has some cost).

Maybe I'm misunderstanding now that I type this. Is `distance` available to `canUse` some way? Or does it have to calculate it itself?

aaclayton: Aah, fair enough. I suppose you're right that you would not have the falling status if your distance to the ground is <= 0. Fair to ignore this comment!

Fyorl: I set a breakpoint and can see that `canUse` is called twice it seems, and the second time seems to be after `preActivate` so we do have `fallDistance`, so I added a check just in case.

aaclayton (approved):

PR #1070, opened 2026-05-21
[#771] Add "important" adversaries, which track Wounds & Madness

### LINKED ISSUE #771: Add the ability to create "important" NPCs that have Wounds and Madness without being a Hero (Player Character)

roth-michael:
Closes #771 
- Adds a checkbox to the header of adversary sheets for Important
- Adversaries no longer delete reserve resources from their schema
- Added a new `usesReserveResources` getter to `CrucibleBaseActor` (defaults to `true`, overridden in `CrucibleAdversaryActor`). This should be used everywhere we previously checked whether wounds/madness _existed_ on an actor
- Moved hero-specific override of `allocateResourceChanges` into parent implementation, with a check for the above getter
- Moved some adversary-specific getter overrides into parent implementation as well

One thing to consider is whether we want to set `woundsMultiplier` and `madnessMultiplier` to 0 for actors which don't use reserve resources. Or clobber bonus & base & max during prep. I think we can just leave it as-is, since having behind-the-scenes wounds values prepped doesn't actually make a difference anywhere with these changes.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -920,6 +920,8 @@
   },
   "ADVANCEMENT": {
     "ExpectedLevel": "Expected Level",
+    "Important": "Important",
+    "ImportantTooltip": "Important Adversaries track Wounds and Madness.",
     "Level": "Level",
     "MilestoneAward": "Award Milestones",
     "MilestoneTooltip": "{progress} of {required} milestones until next level",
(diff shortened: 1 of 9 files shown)


aaclayton (approved):
A couple changes I want to make on top, but we can merge this first.

PR #1086, opened 2026-05-25
Talent fixes

danielrab:
For now, this PR includes
- a fix for Repercussive Block (removing the hook that duplicated the disarm tag, and adding prerequisites checking).
- a small fix to the `canUse` hook of Body Block, making it correctly reject usage when the last action has no events, such as a move action.

Notes:
- Repercussive Block and Body Block have nearly identical `canUse` hooks and nearly identical error messages in those hooks. Perhaps the checks they perform can be standardized into a function that checks that the last action is unconfirmed and has specific tags and outcome, such as a blocked melee attack.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -416,6 +416,11 @@
           "AttackMissed": "You may only Interpose against an attack that hit or glanced your ally.",
           "RequiresStrike": "Interpose requires the most recent action to be a weapon Strike.",
           "SingleTarget": "Interpose can only be used against a single-target Strike."
+        },
+        "REPERCUSSIVE_BLOCK": {
+          "AlreadyConfirmed": "The attack against you has already been confirmed and can no longer be reacted to.",
+          "InvalidOutcome": "You may only use Repercussive Block after an attack against you is defended by Block.",
+          "MeleeOnly": "You may only use Repercussive Block against an incoming melee attack."
         }
       },
       "AuraNoToken": "You cannot use an Aura-target action without having a token on the scene!",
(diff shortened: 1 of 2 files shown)


lang/en.json
```diff
@@ -421,6 +416,11 @@
           "AttackMissed": "You may only Interpose against an attack that hit or glanced your ally.",
           "RequiresStrike": "Interpose requires the most recent action to be a weapon Strike.",
           "SingleTarget": "Interpose can only be used against a single-target Strike."
+        },
+        "REPERCUSSIVE_BLOCK": {
```
aaclayton, line 420: I want to get ahead of an anti-pattern here that I don't like before it goes further. I don't like the idea of adding localization keys that are specific to individual talents. 

We need to generalize these action usage warnings. Something like `AlreadyConfirmed` should be a generic warning message that we can use and reuse whenever it's not possible to use an action because it depends on some prior action to be UNCONFIRMED.

I don't want `BODY_BLOCK` and `REPERCUSSIVE_BLOCK` to both independently declare this almost identical warning. 

Something like `REPERCUSSIVE_BLOCKInvalidOutcome` can instead be a general `WARNINGS.ReactionRequiresBlock`

Something like `REPERCUSSIVE_BLOCK.MeleeOnly` can be `WARNINGS.ReactionRequiresMelee`

etc...

Let's reorganize these i18n keys - we could even change the ones for INTERPOSE/BODY_BLOCK while we are at it if you like.

aaclayton (changes requested):

lang/en.json
aaclayton: Follow-up to this. We should follow the approach of https://github.com/foundryvtt/crucible/pull/1105

aaclayton:
Closed via https://github.com/foundryvtt/crucible/pull/1155

PR #1094, opened 2026-05-26
[#1050] Allow multi-actor skill checks to not show aggregate result

### LINKED ISSUE #1050: Improve the intentionality behind whether a requested skill check is specifically a Group Check or whether it is a set of individual check requests.
Requested skill checks (from enricher) are showing a "Required Successes" counter as though they are group checks. The field should not appear on solo skill checks.

Image

Source: Ember team playtest session 1

roth-michael:
Closes #1050 
Changes:
- `GroupCheck#requestSubmit` now accepts an `options.aggregate` (default `true`) parameter. If `true`, chat card will include aggregate result. If false, will not.
- Single-skill aggregate group checks now flavor their chat cards with "Group {skill} Skill Check."
- `StandardCheckDialog` now includes a checkbox for "Group Check" which defaults to checked for `group` skill check enrichers or programmatically invoked `GroupCheck` dialogs. The checkbox only appears of the Request panel is open and there are at least 2 actors in the list. The checkbox controls whether an aggregate result is displayed.
- `StandardCheckDialog`'s toggles for the Skill Configuration & Request panels are now tied only to configurability, not `isGroupCheck`. This means that a GM can always (unless responding to a requested roll) access these panels, regardless of how the dialog was invoked.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1431,6 +1431,7 @@
     },
     "GROUP_CHECK": {
       "Title": "Group Skill Check",
+      "TitleSpecific": "Group {skill} Skill Check",
       "DialogTitle": "{skill} Check: {name}",
       "NoUserForActor": "No active user found for: {name}",
       "STATUSES": {
@@ -1452,6 +1453,7 @@
       "AnyActorHint": "Choose any world Actors.",
       "ChooseTarget": "Choose Target",
       "ClearRequest": "Clear Request",
+      "GroupCheck": "Group Check",
       "Request": "Request",
       "RequestRolls": "Request Rolls",
       "SpecificActors": "Specific Actors",
--- a/module/dice/group-check.mjs
+++ b/module/dice/group-check.mjs
@@ -7,12 +7,13 @@ import StandardCheck from "./standard-check.mjs";
 
 /**
  * @typedef GroupCheckFlags
- * @property {string} checkId                        Unique group check ID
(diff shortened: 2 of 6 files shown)


aaclayton (approved):
A couple minor nitpicks before we merge.

1. Suppose you click a `[[/skillCheck awareness 13 group]]` where the enricher *explicitly* designates it as a group check. Can we initialize that as `isGroupCheck = true` even if there is < 2 members of the party? I think an explicit enricher `group` designation should require `isGroupCheck = true`.

2. After reducing the requested actors to 1 (or zero), the `this.isGroupCheck` property can remain `true` when it is now practically false. Maybe we should update the `isGroupCheck` property? Or keep the prior value in place? Either way a decision about what happens when we remove actors from a check down to a level where it would - theoretically - become an individual check?

roth-michael:
1. I believe in this case we _do_ initialize `isGroupCheck` as true. Just, when actually performing the check, that is disregarded if there's only one requested actor.
2. Updating it in this context would result in (1) not being the case anymore.

I think `isGroupCheck` should be thought of as "if possible, this will be an aggregate check." It keeps its value behind the scenes so that when you add additional actors, it follows what the enricher (or you, if you removed then added actors) originally defined, but an aggregate check with a single actor isn't a concept which makes sense, I don't think.

aaclayton:
Okay, I'm good with those answers, thanks for the perspective.

PR #1095, opened 2026-05-26
[#1089] Ensure flanking & action-applied status-conveying effects show icon

### LINKED ISSUE #1089: Ensure action-created effects have their `showIcon` set to Always when they convey a status effect or otherwise represent an indefinite but terminable condition
Marking bug since it's a degradation from v13-era instances of such effects.

roth-michael:
Closes #1089 
Truthfully, I think the "best" solution might actually be to override `Token#_drawEffects` - but that's not ideal because we'd have to copy all the code over. Barring implementation of a core `ActiveEffect#shouldBeVisible` getter we can override, this does the job well enough I think.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1151,6 +1151,7 @@ export default class CrucibleActor extends Actor {
       const forceUpdate = effectData._action === "update";
       const shouldUpdate = existing && (reverse ? forceUpdate : !forceDelete);
       const shouldDelete = existing && (reverse ? !forceUpdate : forceDelete);
+      if ( effectData.statuses.length ) effectData.showIcon ??= CONST.ACTIVE_EFFECT_SHOW_ICON.ALWAYS;
       if ( shouldUpdate ) toUpdate.push(effectData);
       else if ( shouldDelete ) toDelete.push(effectData._id);
       else if ( !reverse ) toCreate.push(effectData);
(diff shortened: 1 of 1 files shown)


module/documents/actor.mjs
```diff
@@ -1154,6 +1151,7 @@ export default class CrucibleActor extends Actor {
       const forceUpdate = effectData._action === "update";
       const shouldUpdate = existing && (reverse ? forceUpdate : !forceDelete);
       const shouldDelete = existing && (reverse ? !forceUpdate : forceDelete);
+      if ( effectData.statuses.length ) effectData.showIcon ??= CONST.ACTIVE_EFFECT_SHOW_ICON.ALWAYS;
```
aaclayton: This will throw if it encounters a `{_id, _action: "delete"}` entry

aaclayton (changes requested):
Need at least one bugfix to address a runtime error that would happen when deleting an effect. I am wondering if we should be doing this in `#recordEffectEvents` instead of in `_applyActionEffects()`. Assign a default `showIcon` as ALWAYS if not otherwise defined when the effect is registered to the events stream?

roth-michael:
The issue with doing this in `#recordEffectEvents` is that any effect events added in `postActivate` hooks won't get this treatment. And while sure, `showIcon` could be specified on them since they're programmatic, I think it'd be better if we just get it for free somewhere. Could do it in `#finalizeEvents` in theory. For now, I'm just gonna address the bug you pointed out; if you feel strongly about a better approach I can implement it.

aaclayton (approved):

PR #1096, opened 2026-05-26
[#841, #1047] Add `.scrollable`-style padding, margin, & gutter to creation styles, limit max tooltip height to viewport height

### LINKED ISSUE #841: Allow scrolling tooltips when they go off screen
Currently, when a tooltip is tall enough to overflow a user's screen, it cannot scroll, hiding information from the user.

Example from Ember of the Zeph lineage:
Image

### LINKED ISSUE #1047: Scrollbar overlapping text in ancestry description
On a 2560×1440 monitor at default zoom/scale, the scrollbar in ancestry description panels renders flush against the text with no padding.

Image

Source: Ember team playtest session 1

roth-michael:
Closes #841 
Closes #1047 
For 841, just set `max-height: 100%` on crucible tooltips. Could do something neat like what 5e does with a mutation observer on tooltips to ensure unlocked ones don't render a scrollbar, but instead somehow visually telegraph that there's more info there if you lock it, then if locked scrollbar is rendered. For now, simple solution.

For 1047, I checked how the 5e version in ember looks - it uses the `scrollable` class itself. We could do that too in theory, but at the risk of accidentally introducing sub-styles I just added the important bits to the existing `.step-details` selector.

### DIFF
--- a/styles/creation.less
+++ b/styles/creation.less
@@ -295,9 +295,6 @@
     height: 100%;
     overflow: hidden auto;
     gap: var(--gap);
-    padding-right: 0.75rem;
-    margin-right: -0.75rem;
-    scrollbar-gutter: stable;
 
     .step-header {
       margin: 0;
--- a/styles/creation.less
+++ b/styles/creation.less
@@ -295,6 +295,9 @@
     height: 100%;
     overflow: hidden auto;
     gap: var(--gap);
+    padding-right: 0.75rem;
+    margin-right: -0.75rem;
+    scrollbar-gutter: stable;
 
     .step-header {
       margin: 0;
--- a/styles/theme.less
+++ b/styles/theme.less
@@ -461,6 +461,7 @@
   border-image: url("../ui/frames/tooltip-frame.webp") 64 repeat;
   border-image-width: 32px;
   text-align: left; // always left, regardless of tooltip position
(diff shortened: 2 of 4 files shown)


styles/creation.less
aaclayton, line 293: I think we should ensure that every HTML element that uses `.step-details` that wants it to be scrollable should also assign the `"scrollable"` class. We should get scrolling behavior from `.scrollable` and only add incremental styling rules here on top of `.scrollable` environment, for example:
```css
--scroll-margin: 0.75rem;
```

aaclayton (changes requested):

styles/creation.less
roth-michael: Makes more sense. In this case we don't even need to define additional rules, since 0.75rem is the default and we don't override it anywhere

styles/creation.less
```diff
@@ -295,9 +295,6 @@
     height: 100%;
     overflow: hidden auto;
```
aaclayton, line 296: I believe `overflow` behavior also comes from `.scrollable`

aaclayton (approved):

styles/creation.less
roth-michael: It does, but that was there before I touched it so I figured it maybe had some other effect.

PR #1102, opened 2026-05-27
Ensure effect events are recorded for all relevant targets

### LINKED ISSUE #1160: Ensure effects events are recorded for relevant targets, even if those targets are not being affected by a resource change
e.g. Inspire Heroism

roth-michael:


module/models/action.mjs
```diff
@@ -1694,7 +1691,8 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
       && (SYSTEM.ACTION.TARGET_TYPES[this.target.type]?.region?.ephemeral === false);
     if ( !this.effects.length && !regionEffectRequired ) return;
     const eventsByActor = this.eventsByActor;
-    for ( const [target, events] of eventsByActor ) {
+    for ( const [target] of this.targets ) {
+      const events = eventsByActor.get(target);
```
aaclayton: Swapping to `this.targets` skips over `this.actor` whenever self is not an explicit target, but we may still need to record self-target effects. Probably the cleanest fix is to create an intermediate array for us to iterate over, something like:

```js
const allActors = Array.from(this.targets.keys());
if ( !this.targets.has(this.actor) ) allActors.push(this.actor);
for ( const target of allActors ) {
  ...
```

aaclayton (changes requested):
One change we need to be careful about before merging.

module/models/action.mjs
```diff
@@ -1751,7 +1749,7 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
    * Identify the qualifying CrucibleActionEvent which enables an ActionEffect to be applied.
    * Return true if the effect may be applied unconditionally, or false if qualifications are unmet.
    * @param {CrucibleActor} target              The target actor
-   * @param {ActorEventGroup} events            The pre-classified event group for this target
+   * @param {ActorEventGroup|undefined} events            The pre-classified event group for this target
```
aaclayton: Fix alignment

aaclayton (approved):

aaclayton:
@roth-michael can you please create a github issue for this fix and tag it to `0.9,5` milestone

Issue #1038, opened 2026-05-13
Corrections for a large number of talents that were incorrectly or incompletely implemented


danielrab:
This issue documents all the instances I found where a talent does something different from what its description says or ones that are not fully implemented but lack any disclaimer to indicate that.
There will likely be more. I am slowly working through the talents, seeing what seems to work and what's broken.
I split the list into 6 categories, roughly in the order of urgency:
- those that throw an exception, preventing them from working entirely, or otherwise doesn't work at all when it should.
- those that work but seem to have some issue with how exactly they are implemented.
- those where the implementation looks good, but doesn't match the description
- those that clearly lack a part of the implementation entirely, preventing them from working either at all or partially
- those that work fine but don't check some of the prerequisites for the roll, meaning the GM just needs to keep an eye on what's going on to prevent violations
- typos and other stuff that doesn't really affect gameplay.

The unimplemented talents are only included if the talent lacks a disclaimer that it isn't done like that on Eye of the Storm or Counterspell.

Completely broken:
- [x] Repercussive Block: Exception thrown. Also, doesn't check prerequisites. (addressed in https://github.com/foundryvtt/crucible/pull/1086)
- [x] Evasive Armor: Exception thrown. (addressed in https://github.com/foundryvtt/crucible/pull/1105)
- [x] Still Lake: Exception thrown. (addressed in https://github.com/foundryvtt/crucible/pull/1105)
- [x] Vow of Animus: Silently fails - no effect is being applied, no exception is thrown either. (addressed in https://github.com/foundryvtt/crucible/pull/1102)
- [x] Opportunist: The description seems to refer to mechanics that no longer exist (getting 2 boons on disengagement strikes). Also no automation is provided for the talent, meaning disengagement strikes while flanked remain impossible. It would normally be in a lower category, but this talent can't work without automation at all.
- [x] Evasive Shot: Silently fails.
- [x] Preparedness: Silently fails. (addressed in https://github.com/foundryvtt/crucible/pull/1105)
- [x] Thread the Needle: Silently fails to apply boons.
- [x] Never Yield: Silently fails. (addressed in https://github.com/foundryvtt/crucible/pull/1105)
- [x] Adrenaline Surge: The effect doesn't apply boons. (addressed in https://github.com/foundryvtt/crucible/pull/1103)
- [x] Inspire Heroism: Doesn't apply the effect. (addressed in https://github.com/foundryvtt/crucible/pull/1102)
[...]

PR #1103, opened 2026-05-27
Ensure pre-camel-cased inputs to `generateId` don't get lowercased

roth-michael:
Before & Afters of some examples:
| Input | Before | After |
| ----- | ------ | ----- |
| "Berserker Rage" | "berserkerRage" | "berserkerRage" |
| "berserkerRage" | "berserkerrage" | "berserkerRage" |
| "BERSERKERRAGE" | "berserkerrage" | "bERSERKERRAGE" |
| "BERSERKER RAGE" | "berserkerRage" | "bERSERKERRage" |

The last two rows are, imo, irrelevant edge cases. The important bit is the second row, since we frequently pass camelCase identifiers to `generateId` for the purposes of auto-generating effect IDs, and that shouldn't result in a meaningful change from the identifier.

PR #1105, opened 2026-05-27
[#1038] Many talent fixes

roth-michael:
Addresses most of #1038 
Will be commenting on that issue with the pieces that remain, as well as questions that need answers in order put those pieces together.

**Code Changes**:
- Only store non-permanent summons on an AE, and when such an AE is deleted, delete those non-permanent summons if they are present
- Implement `canUse` for Counter Evade/Riposte/Strike, Defensive Roll, Offhand Strike, Ruthless Momentum
- Ensure Counterspell's `usage.targetAction` is populated early enough that "Recognize Spellcraft" can actually work
- Limit Defensive Roll's movement range
- Ensure Last Stand & Second Wind don't erroneously have dice in their dialog
- Make Last Stand _decrease_ wounds threshold, rather than _increase_
- Add validation to targeting for Executioner's Strike (only viable on targets missing health)
- Implement missing localization for Revive target error
- Automate the passive benefit for Assassin (deadly strikes & spells on unaware targets)
- Automate Berserkers gaining +4 damage on melee strikes while undergoing the Berserker Rage
- Fix bug in damage absorption talents where attacks that did not successfully deal damage would throw errors
- Implement Chirurgeon passive for non-action medicine skill checks in combat
- Restrict Concussive Blows to _melee_ strikes
- Fix Evasive Armor so that it no longer tries to access `total`, which doesn't exist; we now recreate some of the logic used when determining `total` within the hook
- Automate Justiciar (crits targeting health also deal the overflow amount to morale)
- Fix typo in Never Yield hook
- Automate passive benefit for Poisoner (+2 damage whenever dealing Poison)
- Move Preparedness hook to `prepareAction` to avoid erroneous early rejection, fix typo of action id check therein
- Automate Skirmisher benefit (any Reactive Strike gets 2 banes against talent holder)
- Fix bug adding bonus damage to spells in Sorcerer hook
- Fix typo in Still Lake hook
- Ensure `CrucibleActor#_onDealDamage` (and therefore the `applyCriticalEffects` hook) is called for _healing_ as well as damaging actions
- Fix Counterspell bug where rune was set to Lightning and not changeable unless the actor had multiple runes to alternate between
- Automate "determine" inflection in Counterspell flow
- Automate Legerdemain, allowing Tier 2 inflections to be used even when silenced

**Pack Changes**:
- Adrenaline moved from Resources/Action to Resources/Focus
- Correct reference to Spirit rune in Bard
- Have Defensive Roll target self
- Fix missing duration for Dirty Tricks effect
- Fix typo in Dual Wield
- Detail what amount is healed in Nosferatu's Vampiric Bite description
- Added `strong` tags for Toughness & Health in Second Wind (I'm not sure why that stood out to me)
- `An challenging` -> `A challenging` in Stunning Strike
- Removed `skill` & `wilderness` tags from Wildspeaker's Wildspeak, as the talent is intended to be purely informational, not an actual skill attack.


module/hooks/action.mjs
```diff
@@ -176,6 +176,24 @@ HOOKS.bullrush = {
...
 /* -------------------------------------------- */
 
+/**
+ * Shared helper for checking eligibility of an action based on the result of the most recent action
+ * @param {CrucibleAction} action
+ * @param {object} options
+ * @param {"parry"|"block"|"dodge"} options.requiredResult
+ * @throws {Error}
+ */
+function _canUsePostDefend(action, {requiredResult}) {
```
aaclayton, line 186: I don't like using `requiredResult` both as a strict `RESULT_TYPES` key identifier and also as a localization key part. It's trying to be too clever. 

We should pass in a `requiredResult` that is a value in `RESULT_TYPES`, we can then look up the label from `RESULT_TYPE_LABELS`.

module/hooks/talent.mjs
```diff
@@ -132,6 +146,16 @@ HOOKS.battlefocus00000 = {
 
 /* -------------------------------------------- */
 
+HOOKS.berserker0000000 = {
+  prepareAttack(_item, action, _target, rollData) {
+    if ( !this.effects.has(SYSTEM.EFFECTS.getEffectId("berserkerRage")) ) return;
```
aaclayton, line 151: Confirm this is correct with https://github.com/foundryvtt/crucible/pull/1103 now merged?

module/hooks/talent.mjs
```diff
@@ -518,6 +560,26 @@ HOOKS.irrepressiblespi = {
...
 
+HOOKS.justiciar0000000 = {
+  applyCriticalEffects(item, action) {
+    for ( const event of action.events ) {
+      if ( (event.target === this || !event.isCriticalSuccess || !event.damagesHealth) ) continue;
+      const idx = action.events.indexOf(event) + 1;
+      const {overflow, resistance, multiplier} = event.roll.data.damage;
+      const moraleDamage = crucible.api.models.CrucibleAction.computeDamage({overflow, resistance, multiplier});
+      if ( moraleDamage ) {
+        action.recordEvent({
```
aaclayton, line 571: `applyCriticalEffects` happens while iterating events, IIRC. Is it safe to `recordEvent` while iterating events? I think probably, but just having second thoughts here.

module/hooks/talent.mjs
```diff
@@ -821,6 +888,14 @@ HOOKS.saboteur00000000 = {
 
 /* -------------------------------------------- */
 
+HOOKS.skirmisher000000 = {
+  defendAttack(item, action, origin, rollData) {
+    if ( action.id === "reactiveStrike" ) rollData.banes.skirmisher = {label: item.name, number: 2};
```
aaclayton: Code style I'm wanting to move towards is:

```suggestion
    if ( action.id === "reactiveStrike" ) rollData.banes[item.id] = {label: item.name, number: 2};
```

lang/en.json
```diff
@@ -454,6 +452,11 @@
       "MissingSummonEffect": "ActiveEffect data must be defined to track the non-permanent summon created by the action \"{action}\"",
       "MovementTooFar": "The planned movement path exceeds the maximum range allowed by this action.",
       "MovementTooShort": "The planned movement path does not meet the minimum range required by this action.",
+      "MustFollowMainhandStrike": "You may only perform the {action} action after a basic Strike action with your mainhand weapon.",
+      "MustFollowMeleeBlock": "You may only perform the {action} action after Blocking a melee attack.",
+      "MustFollowMeleeDodge": "You may only perform the {action} action after Dodging a melee attack.",
+      "MustFollowMeleeKill": "You ma only perform the {action} action after incapacitating an enemy with a melee attack.",
```
aaclayton:
```suggestion
      "MustFollowMeleeKill": "You may only perform the {action} action after incapacitating an enemy with a melee attack.", 
```

aaclayton (changes requested):
This looks great, effectively approved but a few tiny nitpicks to resolve before merging.

module/hooks/talent.mjs
roth-michael: Yep, #1103 was a requirement for this to work properly in fact.

module/hooks/talent.mjs
roth-michael: I don't think it does happen while iterating. It gets called by `CrucibleActor#_onDealDamage`, which in turn is called only once in `CrucibleAction#confirm`

aaclayton (approved):
Thanks!

PR #1134, opened 2026-06-02
[#1133] Clean up references to ability bonuses, fix defense formulae

### LINKED ISSUE #1133: Remove reference(s?) to "Ability Bonus" as anything except the end composite average, and correct defense formulae in rules journal
I think there's only one spot where we use "Ability Bonuses," but having it as a concept (half the ability score, rounded in the case of a single scaling ability) is needlessly confusing to new users when we can instead present them with a more explicit formula. Example change (and, I think, the one spot):
> Skill Check = 3d8[Boons,Banes] + Average of Ability Bonuses (Round Up) + Skill Bonus + Enchantment Bonus

to

> Skill Check = 3d8[Boons,Banes] + (Sum of Scaling Abilities / 4) (Round Up) + Skill Bonus + Enchantment Bonus

roth-michael:
Closes #1133


aaclayton (approved):
Thanks!

PR #1142, opened 2026-06-02
Make the pictures of actions clickable and their names draggable.

### LINKED ISSUE #1141: Make the pictures of actions clickable and their names draggable.
Currently, only the picture part of an action can be dragged to the hotbar and only their names can be clicked to use the action.
I think both parts should be usable for both purposes.
My players ran into both parts of this at least once.

danielrab:
closes https://github.com/foundryvtt/crucible/issues/1141
note: the tags section remains non-interactive in this implementation. This is on purpose - if I click on a tag I expect either to open its description as a journal or have nothing happen.


aaclayton:
I'm sorry to say I don't actually *want* this change. It was intentional that the name be the clickable handle for using the action while the image be the draggable handle for creating a macro or a journal text link. I don't like how in other systems the entire line-item is draggable so this was an intentional choice. I might be persuaded to change my mind in the future, but for now this is something I want to keep.

PR #1155, opened 2026-06-03
[#1038] Additional talent fixes

### LINKED ISSUE #1038: Corrections for a large number of talents that were incorrectly or incompletely implemented

roth-michael:
Closes #1038 - remaining issues have been (or will be) split out.

Code changes:
- Move `SPECIFIC` i18n into a more generic spot
- Fix Body Block `canUse` hook to properly prevent usage
- Add automation for Evasive Shot
- Limit max movement in Flying Kick to stride
- Remove absolute target validation for Executioner's Strike, instead removing Deadly + Bleeding if target is not under half health
- Amend Interpose `preActivate` to have checks done in `canUse` when possible, in `acquireTargets` when necessary
- Use localization for the "Read X" scroll action name
- Remove unnecessary `postActivate` for Repercussive Block, add `canUse`
- Remove old, non-functional automation for Evasive Shot
- Add automation for Subtle Extrication

Pack changes:
- Amend Counter Evade action cost to -1, not -2
- Amend Counter Riposte description & tags to give Accurate, rather than benefitting from Exposed
- Amend Dustbinder description to reflect half wisdom damage, not full
- Add Evasive Shot action to talent, to be used for the movement
- Add 3-round duration to Executioner's Strike effect (not necessary as it gets merged, but to avoid the `nullS` tag)
- Amend Fan of Arrows to use 40 range, rather than weapon range
- Add movement tag & target type, enemy scope, 1-target limit, 0A cost (uses movement) to Flying Kick action, amend description to reflect change in max range
- Amend Legerdemain description to refer to "Inflections of talent tier 2" rather than "Lesser Inflections"
- Amend Necromancer description to specify half wisdom damage, not full intellect
- Amend Opportunist description to remove reference to non-existent boons
- Amend Pyromancer description to specify half Intellect damage rather than full, and 3 Rounds rather than 1
- Add 1-round duration to Steadfast effect
- Amend Subtle Extrication description to reflect new behavior
- Amend Sundering Strike description to remove reference to mainhand weapon & add reference to Strength-scaling weapon
- Amend Surgeweaver description to specify 3 Round duration rather than 1
- Amend Thoughtbinder description to specify 3 Round duration rather than 1
- Amend Voidcaller description to specify Health damage rather than Morale
- Amend Zealous Smite description to specify Strength-based weapon rather than heavy


module/hooks/action.mjs
```diff
@@ -725,12 +753,12 @@ HOOKS.executionersStrike = {
...
-      const {health} = target.actor.resources;
-      if ( health.value < health.max ) continue;
-      target.error ??= _loc("ACTION.WARNINGS.RequiresTargetWounded", {action: this.name});
-    }
+  preActivate() {
+    const target = this.targets.keys().next().value;
+    const {health} = target.resources;
+    if ( health.value < (health.max / 2) ) return;
+    this.effects = [];
+    this.tags.delete("deadly");
```
aaclayton, line 761: I think `preActivate` is too late to remove deadly - the damage multiplier has already been set. Need to either do it earlier or need to reverse the damage multiplier.

module/hooks/action.mjs
```diff
@@ -429,6 +432,23 @@ HOOKS.electrochargeAmpoule = {
 
 /* -------------------------------------------- */
 
+// TODO: Currently this will consume free movement. Determine how to make that not the case while still properly
+// showing 0 cost on the token ruler
+HOOKS.evasiveShot = {
+  prepare() {
+    this.range.maximum = this.actor.system.movement.stride / 2;
```
aaclayton: Think we need to `Math.round()` this for odd-numbered strides

module/hooks/action.mjs
```diff
@@ -1236,16 +1266,13 @@ HOOKS.reload = {
...
-    for ( const [target, events] of this.eventsByTarget ) {
-      if ( !events.allSuccess ) continue;
-      const {mainhand} = target.equipment.weapons; // TODO - react to the prior action?
-      if ( !mainhand?.id || mainhand.properties.has("natural") ) continue;
-      this.recordEvent({type: "actorUpdate", target,
-        actorUpdates: {items: [{_id: mainhand.id, system: {dropped: true, equipped: false}}]},
-        itemSnapshots: [mainhand.snapshot()],
-        statusText: [{text: "Disarmed!", fontSize: 64}]});
+  canUse() {
+    const targetAction = ChatMessage.implementation.getLastAction();
```
aaclayton, line 1271: Can be undefined.

module/hooks/action.mjs
```diff
@@ -147,22 +147,25 @@ HOOKS.blastFlask = {
 HOOKS.bodyBlock = {
   canUse() {
     const targetAction = ChatMessage.implementation.getLastAction();
```
aaclayton, line 149: Can be undefined.

_source/talent/Sundering_Strike_sunderingstrike0.yml
```diff
@@ -14,14 +14,14 @@ system:
       img: icons/skills/melee/shield-damaged-broken-gold.webp
       condition: ''
       description: >-
-        <p>You make a <strong>Strike</strong> with your mainhand weapon which,
-        if successful, applies the <strong>Exposed </strong>condition to your
-        target for 3 rounds. This attack is <strong>Difficult</strong> to
-        perform and deals reduced damage.</p>
+        <p>You make a <strong>Strike</strong> with your a Strength-scaling
```
aaclayton:
```suggestion
        <p>You make a <strong>Strike</strong> with a Strength-scaling
```

_source/talent/Zealous_Smite_zealoussmite0000.yml
```diff
@@ -3,7 +3,9 @@ type: talent
 _id: zealoussmite0000
 img: icons/magic/fire/dagger-rune-enchant-flame-blue-yellow.webp
 system:
-  description: <p>You can imbue righteous power into your blow with a heavy weapon.</p>
+  description: >-
+    <p>You can imbue righteous power into your blow with a Strength-based
+    weapon.</p>
   actions:
     - id: zealousSmite
       condition: You are attacking an Exposed enemy with a Heavy weapon.
```
aaclayton: Inconsistent

_source/talent/Executioner_s_Strike_executionersstri.yml
aaclayton: This `condition` field is problematic since we don't have a real definition of "Bloodied"

lang/en.json
```diff
@@ -460,10 +450,15 @@
...
       "MustFollowMeleeKill": "You may only perform the {action} action after incapacitating an enemy with a melee attack.",
+      "MustFollowRanged": "You may only perform the {action} action after a ranged weapon attack.",
       "MustFollowStrike": "You may only perform the {action} action after a basic Strike action",
+      "MustReactHit": "You may only perform the {action} action in response to an attack that hit or glanced.",
+      "MustReactSingleTarget": "You may only perform the {action} action in response to an action that had a single target.",
+      "MustReactStrike": "You may only perform the {action} action in response to a weapon Strike.",
       "MustReload": "Your weapon requires reloading in order to use this action.",
       "MustScaleDex": "Every weapon used in this action must scale using dexterity.",
       "MustScaleStrength": "Every weapon used in this action must scale using strength.",
+      "MustTargetTarget": "The {action} action requires that you target the target of the most recent action.",
```
aaclayton: Little clunky wording
```suggestion
      "MustTargetTarget": "The {action} action may only be used on the target of the most recent action.",
```

aaclayton (changes requested):
Couple small remaining callouts

roth-michael:
Pushed up because I've got to step away for the night, but note the change to **Executioner's Strike**'s description. I went with the assumption that the weapon need not be Heavy, but can revert that and implement a check if necessary. Also made it clear that the strike may result in simply a normal Strike, if the opponent is at or above half health.

aaclayton (approved):

PR #1157, opened 2026-06-03
Mark talents & actions that are known not to be implemented as such

roth-michael:


aaclayton (changes requested):
Elsewhere we have consistently used `<h3 class="divider">Playtest Notes</h3>` as a divider/marker for denoting WIP features.

aaclayton (approved):

PR #1167, opened 2026-06-04
add tag enricher

### LINKED ISSUE #1166: Add Rule enricher which allows embedding a tooltipped inline element that describes a specific game mechanic
There's currently a `@Condition` enricher, which can display conditions inline, there should also be a similar enricher for displaying weapon tags inline as those appear in descriptions of actions quite often and the players can't be assumed to know all the tags.

danielrab:
closes https://github.com/foundryvtt/crucible/issues/1166

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -577,6 +577,7 @@ function enrichCondition([match, conditionId]) {
   tag.innerHTML = _loc(cfg.name);
   tag.dataset.crucibleTooltip = "condition";
   tag.dataset.condition = conditionId;
+  tag.classList.add("inline-enriched");
   tag.classList.add("condition");
   return tag;
 }
@@ -597,6 +598,7 @@ function enrichActionTag([match, tagId]) {
   tag.innerHTML = cfg.label;
   tag.dataset.crucibleTooltip = "tag";
   tag.dataset.tag = tagId;
+  tag.classList.add("inline-enriched");
   tag.classList.add("tag");
   return tag;
 }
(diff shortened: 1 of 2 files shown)


module/enrichers.mjs
```diff
@@ -55,6 +55,11 @@ export function registerEnrichers() {
       pattern: /@Condition\[(\w+)]/g,
       enricher: enrichCondition
     },
+    {
+      id: "crucibleTag",
+      pattern: /@Tag\[(\w+)]/g,
+      enricher: enrichTag
```
aaclayton: While there is probably only one concept of "Tag" in the crucible system, I think I'd prefer calling this "ActionTag" for the sake of clarity and future-proofing.

```suggestion
      id: "crucibleActionTag",
      pattern: /@ActionTag\[(\w+)]/g,
      enricher: enrichActionTag
```

module/enrichers.mjs
```diff
@@ -576,6 +581,26 @@ function enrichCondition([match, conditionId]) {
...
+/**
+ * Enrich a condition reference into an interactive element displaying the tag name and tooltip.
+ * @param {RegExpMatchArray} matchArray
+ */
+function enrichTag([match, tagId]) {
+  console.log(tagId);
+  const cfg = SYSTEM.ACTION.TAGS[tagId];
+  if ( !cfg ) return new Text(match);
+  const tag = document.createElement("enriched-content");
+  tag.innerHTML = cfg.label;
```
aaclayton, line 598: Do we want an icon here as we do with other enricher inline embeds?

aaclayton (changes requested):

danielrab:
I realized some things that should have been done differently, so I'll just open a new PR

PR #1170, opened 2026-06-04
More condition enrichers in talents

### LINKED ISSUE #1169: Many talents refer to conditions with just bold text instead of an enricher.

danielrab:
closes https://github.com/foundryvtt/crucible/issues/1169
This fixes all instances that I found of a talent or its action mentioning a condition as bold text rather than enricher.


aaclayton (approved):
Good improvements, thanks!

PR #1179, opened 2026-06-07
uncap resources

### LINKED ISSUE #1178: Remove schema-enforced caps on `resource.value` fields in favor of dynamic clamping on the value of the maximum field.
steps to reproduce:
- create an actor
- create an effect on that actor
- add a change to that effect adding a big number to `system.resources.action.bonus` or an equivalent for focus or heroism
- try to set the resource to a big value

expected result: the resource is set to the big value
observed result: while the maximum reported on the sheet is successfully increased, but trying to set the actual value of the resource caps it to a lower value

danielrab:
closes https://github.com/foundryvtt/crucible/issues/1178

### DIFF
--- a/module/const/attributes.mjs
+++ b/module/const/attributes.mjs
@@ -217,22 +217,19 @@ export const RESOURCES = defineEnum({
     label: "RESOURCES.Action",
     tooltip: "RESOURCES.TOOLTIPS.Action",
     type: "active",
-    color: Color.from("#FF9900"),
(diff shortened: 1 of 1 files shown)


aaclayton:
I took a bit of a wider approach to this improvement. Closed via https://github.com/foundryvtt/crucible/commit/47137d6a082bf6e8b8975b2968412a9b9d9cac5d

PR #1191, opened 2026-06-08
[WIP] more enrichers

danielrab:
a replacement for #1167


danielrab:
now this once again closes #1166

danielrab:
At this point, all the enrichers that I wanted to implement are implemented.
We may want to refine some of the tooltips, and we definitely will be adding more stuff to the `@Rule` enricher, but neither of those affects our ability to start using the enrichers in the item descriptions and journals, so those changes can come in a later PR, so I think we can merge this if you don't see some part of what I've done so far that you'd prefer to be done differently.

danielrab:
Obsolete now

PR #1193, opened 2026-06-08
[#1109] Ensure all necessary roll data is present prior to calling hooks

### LINKED ISSUE #1109: Ensure target-based boons/banes are available in the `prepareAttack` actor hook
Pack Hunter: We access `rollData.boons.flanked.number` in the `prepareAttack` hook, but we actually add the flanked boons in `_configureTargetRollData` which is called _after_ `_configureAttackerRollData`. Probably we want to move the hook calling for both `prepareAttack` and `defendAttack` out of the above functions, calling them only after both are done.

roth-michael:
Closes #1109 
Moved `prepareAttack` and `defendAttack` hook calls out until _after_ the base roll data for attacker & defender are resolved. This ensures things like Pack Hunter (which is a property of the attacker but depends on the flanked status of the defender) work properly. The only time both attacker & target configuration weren't called back to back was for hazards, but for hazards the `defendAttack` hook wasn't called anyway, so this was nondisruptive. 

I'd be open to making `_configureRollData` non-static and just not passing `actor`, instead using `this`, but it felt appropriate.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -344,11 +344,10 @@ export default class CrucibleActor extends Actor {
   /**
    * Configure attack roll options contributed by the acting actor based on their status conditions.
    * @param {CrucibleAction} action                 The action being performed
-   * @param {CrucibleActor} target                  The target actor being attacked
    * @param {AttackRollData} rollData       The mutable roll data for the attack
    * @internal
    */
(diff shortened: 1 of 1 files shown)


aaclayton (approved):
Looks clean to me.

PR #1202, opened 2026-06-09
add tooltip to weapon category

### LINKED ISSUE #1139: Surface weapon scaling ability on the weapon sheet
**Problem**
Actions with a Finesse or Brute requirement (e.g. Weak Points, Lunge) call for "a weapon which scales using Dexterity/Strength," but a weapon's scaling ability isn't shown anywhere on its item sheet. Scaling is set by the weapon's category (`module/const/weapon.mjs`) and is documented in System Rules → Equipment ("Weapon Categories" table) — but on the weapon itself only the category name is visible. To tell whether a given weapon satisfies a Finesse/Brute action, you have to read its category and cross-reference the Equipment rules table from memory.

**Steps to reproduce**
1. Open any weapon item sheet (e.g. a Light or Heavy weapon).
2. The sheet shows Category, quality, damage type, properties — but not the scaling ability.
3. Open a Finesse action (e.g. Weak Points): its tooltip requires a Dexterity-scaling weapon, but nothing on the weapon confirms whether it qualifies.

**Expected**
The weapon sheet should communicate its scaling ability (and/or whether it counts as Finesse/Brute), so action prerequisites are legible without leaving the sheet.

**Suggested direction (for discussion)**
The value is already derived at `system.config.category.scaling`. Two in-pattern options:
- A derived tag chip in the item header (`templates/sheets/item/item-header.hbs` already renders `crucibleTags`) — e.g. a STR / DEX (or Finesse / Brute) chip, reusing the Combat-rules vocabulary.
- A read-only "Scaling" line on the weapon config tab next to Category (`templates/sheets/item/weapon-config.hbs`).

Either or both; the header chip is the more visible.

**Environment**
Crucible 0.9.4, Foundry VTT 14.363.

(Per CONTRIBUTING: flagging interest in working on this if it's approved for community work — not starting code yet.)

danielrab:
closes #1139 
example tooltips added by this PR:
image
image
image
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2650,6 +2650,13 @@
       "ReloadTooltip": "This weapon requires reloading in order to strike. When the weapon is not loaded, any actions which use the weapon cost +1 Action Points.",
       "Returning": "Returning",
       "ReturningTooltip": "This weapon is magically or mechanically engineered to return to your hand after being thrown. Using the Throw Weapon action with this weapon does not result in it becoming Dropped.",
+      "TrainedTooltip": "This weapon uses {training} training and scales with {scaling}. This weapon needs {hands} to use.",
+      "TrainedTooltipHands": {
+        "one": "one hand",
+        "both": "both hands",
+        "main": "the main hand",
+        "offhand": "the offhand"
+      },
       "Thrown": "Thrown",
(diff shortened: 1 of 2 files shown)


danielrab:
I'll be taking parts of this PR into #1191 and then reopening it with a compatible implementation.

PR #1238, opened 2026-06-12
[#1116] Ensure hazard & roll dialogs are properly styled in firefox

### LINKED ISSUE #1116: Request panel displays incorrectly in Firefox
Image

roth-michael:
Closes #1116 
Three parts to this.
1. Unrelated to the issue, now when we toggle the left panel of the hazard or roll dialog, the rest of the dialog stays in the same relative spot (adjust `left` by what the width will change by).
2. The pips wrapping like crazy in firefox: Now by default `.crucible-rank-pips` have `flex-wrap: nowrap`, and we change it in the one spot that we actually want wrapping: the actor sheet styles.
3. The unexpected wrapping of paneled dialogs & unexpectedly _loooong_ hazard dialog: Hardcoded width for standard check without panels, standard check with at least one panel, standard check with both possible panels. I hate it, because it feels like it _should be automatic_, and indeed in chrome it is. I hate it, because if we ever change the base size of the panels, we need to change these styles too. But it is a solution to the firefox css issue, and it plays nice with font size scaling, so I think it's better than letting the bug exist.

### DIFF
--- a/module/dice/hazard-dialog.mjs
+++ b/module/dice/hazard-dialog.mjs
@@ -241,7 +241,9 @@ export default class HazardDialog extends ActionUseDialog {
    */
   static async #onHazardConfigToggle(_event) {
     this.#configureHazard = !this.#configureHazard;
-    await this.render({window: {title: this.title}});
+    const remInPx = parseFloat(getComputedStyle(document.documentElement).fontSize);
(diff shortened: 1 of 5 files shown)


aaclayton:
Fixed by switching to grid https://github.com/foundryvtt/crucible/commit/632f2d466e615c05cd1ce671a6e0ebcdad4a6706

PR #1251, opened 2026-06-13
Add Rule enricher which allows embedding a tooltipped inline element that describes a specific game mechanic

danielrab:
As was requested, I am splitting #1191 into smaller PRs, this is the first one, it includes the enricher itself, some CSS changes, data for the action tag version of this enricher.
As a way of demonstrating how it all works together, this PR addresses the erroneous usage of `@Condition[weakened]` mentioned in #1166, replacing them with `@Rule[action.weakened]`


module/const/system.mjs
```diff
@@ -203,6 +203,10 @@ export const TIME = Object.freeze({
   restSeconds: 60 * 60 * 10
 });
 
+export const RULES = {
```
aaclayton, line 215: We need a clearly documented `@type` or `@typedef` that articulates exactly what data spec is necessary in order for a rule to be valid.

module/enrichers.mjs
```diff
@@ -623,11 +628,30 @@ function enrichSpell([match, spellId]) {
...
+ * Enrich a rule reference into an interactive element displaying the rule name and tooltip.
+ * @param {RegExpMatchArray} matchArray
+ */
+function ruleEnricher([match, tagId, label]) {
+  const cfg = foundry.utils.getProperty(SYSTEM.RULES, tagId.replaceAll("-", "_"));
+  if ( !cfg ) return new Text(match);
+  const tag = document.createElement("enriched-content");
+  tag.innerHTML = label ?? _loc(cfg.label);
+  tag.dataset.crucibleTooltip = "tag";
+  tag.dataset.crucibleTooltipText = _loc(cfg.tooltip, {passiveBase: SYSTEM.PASSIVE_BASE}) ?? `[WIP] no tooltip was provided for ${cfg.label}`;
```
aaclayton: There shouldn't be anything specific to `SYSTEM.PASSIVE_BASE` here. If a tooltip needs the passive base, it should be pre-encoded in the `cfg.tooltip`.

At a more general level, should probably require that `rule.tooltip` values must be pre-localized rather than being localized here. Many rule tooltips will require formatting, and we can't know here in the general rule enricher what formatting parameters are necessary.

aaclayton (changes requested):
Looks good, two important changes though before we can merge this in!

module/enrichers.mjs
danielrab: oh, yeah, that passiveBase is a leftover from something I was trying to do in the bigger version of this. And yeah, prelocalized sounds good.

module/const/system.mjs
danielrab: this has been done

danielrab:
alright, it now can load a journal page. It also happens to be able to fully emulate the `@Condition` enricher

module/const/system.mjs
```diff
@@ -205,6 +206,19 @@ export const TIME = Object.freeze({
 
 /* -------------------------------------------- */
 
+/**
+ * @typedef {({label: string}|{name: string})&({tooltip:string}|{page:string})} RuleItem
```
aaclayton: 1. I don't like supporting both `name` and `label`. Let's pick one and require it. `name` is probably better but I'm okay either way

2. I think that potentially only a `page` is required because we could get a name from `page.name`. Maybe a simpler way to document this at the expense if a slight amount of precision would be just to list everything as optional and explain in the sentence below that at least some combination of properties is required.
```suggestion
 * @typedef {{name?: string, tooltip?: string, page?: string}} RuleItem
```

module/interaction.mjs
```diff
@@ -203,12 +203,18 @@ async function displayCondition(event) {
  */
 async function displayTagTooltip(event) {
   const element = event.target;
-  const tooltip = element.dataset.crucibleTooltipText ?? SYSTEM.ACTION.TAGS[element.dataset.tag]?.tooltip;
+  let tooltip = element.dataset.crucibleTooltipText ?? SYSTEM.ACTION.TAGS[element.dataset.tag]?.tooltip;
+  let name = element.innerText;
+  const cfg = foundry.utils.getProperty(SYSTEM.RULES, element.dataset.ruleId);
+  if (cfg) {
+    tooltip ??= cfg.tooltip ?? (await fromUuid(cfg.page))?.text.content;
+    name = _loc(cfg.label) ?? _loc(cfg.name);
```
aaclayton: Let's allow getting the `name` from `page.name` if page is specified and `name` is not.

aaclayton (changes requested):

module/const/system.mjs
danielrab: as discussed in Discord, the support for label is for compatibility with existing items

module/interaction.mjs
danielrab: done

Issue #1273, opened 2026-06-16
when a character has both Block and Parry, physical defense result is calculated incorrectly


danielrab:
I'll explain it properly in the upcoming PR, it's easier to explain with the code at hand

PR #1276, opened 2026-06-16
fix physical defense calculation

danielrab:
usage of `dodge` instead of `parry` in the calculation of `block` constant resulted in the a portion of Block up to the value of Parry essentially being converted to Armor


danielrab:
this closes #1273

aaclayton (approved):
oops!

PR #1303, opened 2026-06-17
[#1270] Localize default 'DC' label for unknown defense type

### LINKED ISSUE #1270: Missing "DC" abbreviation translation
“DC” is not translated in chat messages.

Image

roth-michael:
Closes #1270 
Had to make it a getter to be able to localize. Also, used it in the attack roll class instead of hardcoding the localization there again.


aaclayton (changes requested):
I don't really like this getter approach. We don't use `DEFENSE_TYPE` anywhere except one place. Let's drop it entirely and just localize in-place when we need it?

roth-michael:
We technically use it in two spots, but I'm happy to drop it.

aaclayton (approved):

PR #1304, opened 2026-06-17
[#750] Enable locking talent descriptions in tree via middle-mouse

### LINKED ISSUE #750: Consider allowing previewed Talents in the talent tree to somehow be "locked," as with tooltips, to allow for hovering over nested action tags
Would have to consider the UX for telling the user that the "tooltip" is locked. Which is also a consideration we should probably make for _normal_ crucible tooltips.

roth-michael:
Closes #750

I expect there may be some refinement of this mechanism & styling, but currently:
- Middle mouse locks the hovered talent's description
- Locked talent descriptions have a box shadow highlight
- While a talent is locked, hover events are skipped for other Talents or Nodes
- Left clicking any other talent, any other node, or any blank space on the tree unlocks, but does not perform the left click interaction
- Left clicking the locked talent unlocks & performs the left click interaction
- Right clicking does nothing, but right click dragging & scrolling is still possible with a description locked

I did _not_ implement locking _node_ tooltips, but maybe we should just for consistency.

### DIFF
--- a/module/canvas/tree/talent-tree-node.mjs
+++ b/module/canvas/tree/talent-tree-node.mjs
@@ -116,7 +116,7 @@ export default class CrucibleTalentTreeNode extends CrucibleTalentIcon {
 
     // If a talent's description is locked, unlock it & pseudo-hover rather than (de)activating node
     if ( tree.hud.target?.isLocked ) {
-      tree.hud.target.toggleLock(false);
+      tree.hud.target.toggleLock();
       this._onPointerOver(event);
       return;
     }
--- a/module/canvas/tree/talent-tree-node.mjs
+++ b/module/canvas/tree/talent-tree-node.mjs
@@ -113,10 +113,17 @@ export default class CrucibleTalentTreeNode extends CrucibleTalentIcon {
     event.stopPropagation();
     if ( event.data.originalEvent.button !== 0 ) return; // Only support standard left-click
     const tree = game.system.tree;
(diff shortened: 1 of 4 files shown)


module/canvas/tree/talent-tree-node.mjs
```diff
@@ -113,10 +113,17 @@ export default class CrucibleTalentTreeNode extends CrucibleTalentIcon {
     event.stopPropagation();
     if ( event.data.originalEvent.button !== 0 ) return; // Only support standard left-click
     const tree = game.system.tree;
-    const ownedTalents = tree.actor.system.talentNodes[this.node.id] || [];
-    const nodeTalents = new Set([...this.node.talents, ...ownedTalents]);
+
+    // If a talent's description is locked, unlock it & pseudo-hover rather than (de)activating node
+    if ( tree.hud.target?.isLocked ) {
+      tree.hud.target.toggleLock(false);
```
aaclayton: `toggleLock` takes no parameter, should either add a force-enable parameter or drop it here.
```suggestion
      tree.hud.target.toggleLock();
```

aaclayton (approved):

module/canvas/tree/talent-tree-node.mjs
roth-michael: Ah yeah, that was some vestigial code. Originally I had it take a `state` parameter, but then realized I was only ever calling it with `false` for checks behind `isLocked`, so I got rid of it. Good catch.

PR #1330, opened 2026-06-19
Align sheet `TABS` with core `ApplicationV2` spec

### LINKED ISSUE #1331: Update crucible sheets' `TABS` property to align with `ApplicationV2`'s specification

roth-michael:
Everywhere we define `static TABS`, move away from our custom `Record<string, Array<Record<string, ApplicationTab>>>` handling to the standard `Record<string, ApplicationTabsConfiguration>` handling. Doing so gives us inferred `group`s (since all of our apps only have one tab group), lets us make use of `labelPrefix` to save writing out labels in most cases, lets us remove what is effectively duplication of `ApplicationSheetV2#_prepareTabs`'s logic, and (perhaps most importantly) gives ChaosOS something pretty to point at as an example.

One change worth calling out: `CrucibleBaseItemSheet#_prepareTabs` now exists as an extension. It assumes `tabs.description` exists (in other words: It assumes `group` will always be `"sheet"`. This is a safe assumption because `SomeSheet#_prepareTabs` is only called in `SomeSheet#_prepareContext`, and in our case we never call `super._prepareContext`; rather, we _directly_ call `_prepareTabs("sheet")`.

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -219,11 +219,11 @@
       "Title": "Aktion"
     },
     "TABS": {
-      "Description": "Beschreibung",
-      "Effects": "Effekte",
-      "Hooks": "Hooks",
-      "Target": "Ziel",
-      "Usage": "Verwendung"
+      "description": "Beschreibung",
+      "effects": "Effekte",
+      "hooks": "Hooks",
+      "target": "Ziel",
+      "usage": "Verwendung"
     },
     "TAG": {
       "Accurate": "Präzise",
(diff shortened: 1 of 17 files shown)


aaclayton (approved):
This makes sense to me. For a bit of history here, I started doing `static TABS` in Crucible before it was an option in core. The core implementation came as an improvement of this approach, so it doesn't surprise me that the crucible solution lags behind the state of the art some. Thanks for these fixes!

PR #1332, opened 2026-06-19
[#1326] Add automation for Iramancer's bonus damage

### LINKED ISSUE #1326: The Iramancer does not deal damage with spell attacks
The character is in the Enraged state from Berserk, yet spells do not receive a +2 damage bonus.

Image

Image

For more information about Gesture Strike, which also covers this, click here:
https://discord.com/channels/170995199584108546/1116696333894570094/1517299488908378133

roth-michael:
Let's get this vacation party started
Closes #1326

### DIFF
--- a/module/hooks/talent.mjs
+++ b/module/hooks/talent.mjs
@@ -848,8 +848,7 @@ HOOKS.intellectualsupe = {
 
 HOOKS.iramancer0000000 = {
   prepareAttack(_item, action, _target, rollData) {
-    if ( !["spell"].some(t => action.tags.has(t)) || !this.statuses.has("enraged") ) return;
-    rollData.damageBonus += 2;
+    if ( action.tags.has("spell") && this.statuses.has("enraged") ) rollData.damageBonus += 2;    
   }
 };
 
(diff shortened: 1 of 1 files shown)


module/hooks/talent.mjs
```diff
@@ -846,6 +846,15 @@ HOOKS.intellectualsupe = {
 
 /* -------------------------------------------- */
 
+HOOKS.iramancer0000000 = {
+  prepareAttack(_item, action, _target, rollData) {
+    if ( !["spell"].some(t => action.tags.has(t)) || !this.statuses.has("enraged") ) return;
+    rollData.damageBonus += 2;
```
aaclayton: We don't really benefit from an early-return here, since we're only doing one thing, so code style here would probably be better in the affirmative:

```suggestion
    if ( action.tags.has("spell") && this.statuses.has("enraged") ) rollData.damageBonus += 2;    
```

aaclayton (changes requested):

module/hooks/talent.mjs
roth-michael: Oof, yeah. Double the same with that needless `[oneThing].some` call.

aaclayton (approved):

PR #1338, opened 2026-06-20
Ensure best-ranked weapon is selected, fix unavailable `weaponChoice` sticking on window close

roth-michael:
`action.mjs` fix was a typo - `_getWeaponAvailability` returns an object, not just the rank.
Fix in the use dialog is so that selecting an UNAVAILABLE weapon & then closing the window doesn't try to "lock" the UNAVAILABLE weapon when you go to use a Strike next.


aaclayton (approved):
Good fixes thanks!

PR #1341, opened 2026-06-22
[#1242] Query maintenance dialog, fix `CrucibleActor#getDesignatedUser`

### LINKED ISSUE #1242: Maintenance dialog should be socketed to the designated owner of the actor
While they still exist, at any rate. Right now it's just prompted for the GM. Should be a query to the designated owner.

roth-michael:
Closes #1242 
Not null-checking `this.getDesignatedUser()` because I _know_ there will be an active GM at least, otherwise this code wouldn't be running. 

Fixes to the `getDesignatedUser` function are
1. Typing - would be `null` not `undefined`
2. An earlier `return true` doesn't matter; if any user returns `true` they'll be the designated user. So if we want to prioritize (i.e. user with this actor assigned, but if no such user then any player owner) then we need to do it separately.

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -605,12 +605,12 @@ export default class CrucibleActor extends Actor {
   /**
    * Find the best active user to roll on this actor's behalf, preferring an assigned character match then any
    * owner, falling back to the active GM.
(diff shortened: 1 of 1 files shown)


aaclayton (approved):
Nice! I thought this would be more complicated.

PR #1342, opened 2026-06-22
[#1278] Ensure description doesn't populate an unnecessary scrollbar

### LINKED ISSUE #1278: Scrollbar erroneously showing for action description in chat/action use dialog with certain scalings
First observed with UI configured so that Font Size is `4` rather than default `5`, but also observed when adjusting browser scaling with `ctrl`+`+`, for instance (and indeed, Font Size `4` and 110% zoom "fixes" as well). 

This is to do with `line-height` on the description `div`. Simple fix is to simply add `line-height: normal` to `.crucible.action-roll .description`, but perhaps there's a better solution out there.

Image

roth-michael:
Closes #1278 
Not sure if maybe there's some secret sauce we should be using instead of this but it feels like a fairly harmless solution. Before & after pics below, showing both a valid "needs to scroll" box on the left and an invalid "no need to scroll" box on the right.

**With font size 4**:
Before: 
image
After: 
image

**With font size 5 (default)**:
Before: 
image
After: 
image

### DIFF
--- a/styles/dice.less
+++ b/styles/dice.less
@@ -453,7 +453,7 @@
   .description {
     max-height: 200px;
     overflow-y: auto;
-    line-height: normal;
+    line-height: var(--font-size-16);
   }
 }
 
--- a/styles/dice.less
+++ b/styles/dice.less
@@ -453,6 +453,7 @@
   .description {
     max-height: 200px;
     overflow-y: auto;
+    line-height: normal;
   }
 }
 


styles/dice.less
```diff
@@ -453,6 +453,7 @@
   .description {
     max-height: 200px;
     overflow-y: auto;
+    line-height: normal;
```
aaclayton: I think instead of `normal` we should probably set an explicit font size and line-height here. I *think* the problem is because we use font-size 14px here with a line height of 1.25em which is produces a non-integer result. Maybe try:
```suggestion
    line-height: var(--font-size-18);
```

aaclayton (changes requested):

styles/dice.less
roth-michael: Works. Size 18 is a bit roomy though. 16 looks better to me. Thoughts?
16 vs 18 side-by-side:
image

aaclayton (approved):

PR #1343, opened 2026-06-22
[#1284] Reduce `freeHands` and `spellHands` by 1 if grappling

### LINKED ISSUE #1284: It is possible to fill both hands with weapons while grappling an opponent
1. Have 1h weapon equipped
2. Successfully grapple enemy
3. Equip shield

You're now grappling with both hands full. We ought to prevent equipping a weapon if it would take up the second hand while grappling.

roth-michael:
Closes #1284 
Dunno if there's a better spot to put this. If you disagree with the logic, let me know; maybe I'm making an assumption here and after beginning the grapple one could get them locked up using their legs, or something.

### DIFF
--- a/module/models/actor-base.mjs
+++ b/module/models/actor-base.mjs
@@ -714,6 +714,12 @@ export default class CrucibleBaseActor extends foundry.abstract.TypeDataModel {
     // Two-handed weapons allow one hand free for spellcasting or single-hand actions
     if ( weapons.twoHanded && !mhTalisman ) weapons.freeHands = weapons.spellHands = 1;
 
+    // Being the aggressor in a Grapple must occupy one hand
+    if ( this.parent.effects.has(crucible.api.hooks.action.grapple._GRAPPLING_EFFECT_ID) ) {
+      weapons.freeHands = Math.max(0, weapons.freeHands - 1);
+      weapons.spellHands = Math.max(0, weapons.spellHands - 1);
+    }
+
     // Multi weapon properties
     weapons.dualWield = weapons.unarmed || (mh?.id && oh?.id && !weapons.shield);
     weapons.dualMelee = weapons.dualWield && !mhCategory.ranged && !ohCategory.ranged;


aaclayton (changes requested):
This is an okay solution, but I don't like encoding a specialized awareness on a specific talent/action like this. I think I'd like to zoom out a step and explore a more generalized solution.

We have a concept of "maintenance" on Active Effects already. I think continued use of one or more hands (or spellHands) should perhaps be a component of the maintenance requirement for an active effect. It requires one hand to maintain a grapple. 

Our equipment preparation logic would then more generally traverse current active effects to track how many hands are being reserved, opening the door for other future actions that require similar treatment.

This becomes a somewhat larger change, and introduces the need to handle certain corner cases like whether some change to available hands would cause a maintained effect to forcibly expire and the UX around that.

Do you have an appetite to take that on? Would you want to push this in for now with a FIXME so that we solve the problem temporarily in a hacky way and take a follow-up to implement a more robust solution? Something else?

roth-michael:
I think tracking "spent" hands on effects makes sense. Whether it makes sense _on maintenance_ I'm less sure about; maintenance is currently rather explicitly a repeated cost. We could change that, but I think having it represent something which must explicitly be done each turn to maintain, rather than something which simply _is passively true while the effect exists_, makes sense to me.

If done properly, I think edge cases would be minimized; you wouldn't be able to do something willingly which reduces your available hands further, so we would only need to worry about external effects reducing hands; which isn't something that we currently account for, I don't think.

I don't think this needs imminent fixing, so I'd be happy to close this and work on a more generic handling once we iron out what we want it to look like. The only times having this un-fixed will prove an issue (ignoring testing) are:
1. A weapon-user grapples, then for some reason equips another weapon/shield while grappling
2. A spellcaster who also grapples foes and is able to use two-handed gestures while grappling

I expect both of these to be rare enough that we don't need to merge the current fix, even if it means the bug remains through the next patch.

aaclayton:
Let's close this then and convert create a feature issue to track hand requirements on Active Effects.

PR #1347, opened 2026-06-22
[#1285, #1293, #1294] Fix various bomb misconfigurations

roth-michael:
Closes #1285 by changing effect scope on each bomb from "Enemies" to "All"
Closes #1293 by adding Fire & Acid tags, specifying that type of damage in their description, and fixing the mentioned typo
Closes #1294 by addressing each point in the issue

PR #1348, opened 2026-06-22
[#1299] Fix `relativeTo` for various enrichment calls, amend gem descriptions

### LINKED ISSUE #1299: Gem of Conjured Flame's use of `@ref` means its chat description is incorrect
Image

roth-michael:
Closes #1299 
Pack changes:
- Gem of Conjured Flame now uses `@ref[item.name]` instead of `@ref[name]`
- Gem of Illumination update to use current version of Luminous affix

Code changes:
- In order to support being able to use `@ref[item.name]` on an action, each time we enrich an action's description it must use itself as the `relativeTo`, where previously it was sometimes using the item it was on. So, each time we enrich actions in a loop, I've ensured we're doing so with `relativeTo` set to the action itself.
- I did the same for actions granted by affixes on the affix sheet, and for _affixes_ on an item on the _item_ sheet.
- Finally, an Action's `item` was pointing at an active effect document in the case that it was granted via affix; I've modified the logic in `CrucibleAction#_configure` such that, if it is indeed an affix-riding action, it uses the affix's `item` property to populate `item`. This means that affixes which reference `@ref[item.name]` will work properly now.

There are additional actions which currently use `@ref[name]` to reference their item's name (which has gone unnoticed because the action shares a name with the item) that I'd like to change, but those should wait until #1347 goes in, as all of the bombs do it. The full list of items using `@ref[name]` when it should be `@ref[item.name]` (which will be a follow-up PR):
- Alchemist's Fire
- Antitoxin
- Blast Flask
- Caustic Phial
- Choking Ampoule
- Electrocharge Ampoule
- Frostdrop Flask/Phial
- Lantern
- Paralytic Vial's "Ingest Paralytic"
- Poison Vial's "Ingest Poison"
- Torch

### DIFF
--- a/_source/equipment/Gem_Of_Conjured_Flame_gemConjuredFlame.yml
+++ b/_source/equipment/Gem_Of_Conjured_Flame_gemConjuredFlame.yml
@@ -27,10 +27,10 @@ system:
       name: Summon Flame Elemental
       img: icons/commodities/gems/gem-rough-cushion-orange-red.webp
       description: >-
-        <p>The @ref[name] shatters in a burst of flame, leaving in its place a
-        fiery elemental. Elementals summoned by this gem remain for 6 Rounds (1
-        minute).</p><ul><li><p><strong>Shoddy - </strong>No elemental is
-        summoned</p></li><li><p><strong>Standard </strong>-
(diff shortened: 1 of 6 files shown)


aaclayton (approved):
Checks out from my POV, thanks.

PR #1349, opened 2026-06-22
[#1306] Pass `actionId.itemId` to macro drag data for consumable-bound actions

### LINKED ISSUE #1306: Dragging a consumable-granted action to the hotbar will result in a nonfunctional macro
Consumable-bound actions are prepared as `{actionId}.{itemId}` in `CrucibleBaseActor##registerItemActions`, but `Action#toMacroDragData` passes only the action ID.

roth-michael:
Closes #1306

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -3066,6 +3066,7 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
    * @returns {{type: "crucible.action", macroData: object}}
    */
   toMacroDragData() {
+    const amendedId = this.item?.type === "consumable" ? `${this.id}.${this.item.id}` : this.id;
     return {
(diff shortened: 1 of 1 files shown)


module/models/action.mjs
```diff
@@ -3066,14 +3066,15 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
    * @returns {{type: "crucible.action", macroData: object}}
    */
   toMacroDragData() {
+    const amendedId = this.item?.type === "consumable" ? `${this.id}.${this.item.id}` : this.id;
```
aaclayton, line 3069: Fine for now, but maybe this should be a getter on the action since I'm sure we do this in more than one place.

aaclayton (approved):

PR #1350, opened 2026-06-22
[#1322] Fix NaN-displayed percentages for physical defenses on sheet, ensure minimum per physical defense is 0

### LINKED ISSUE #1322: Incorrect armor designation for creatures without Armor defense
For example, let's say we have a creature with only the Dodge defense. If we set its HP to 0, the values become corrupted and “NaN” is displayed.

Image

Image

In addition, @danielrab applied an effect that reduced the armor value of an unarmored character, resulting in negative values.

Image

roth-michael:
Closes #1322 
If total physical defense is 0, treat it as 1 purely for the purposes of determining a percentage contribution of each (0, mandatorily) physical defense.
Additionally, during final defense prep, cap each individual total (armor, dodge, parry, block) to a minimum of 0, preventing negative totals.

### DIFF
--- a/module/applications/sheets/base-actor-sheet.mjs
+++ b/module/applications/sheets/base-actor-sheet.mjs
@@ -254,6 +254,7 @@ export default class CrucibleBaseActorSheet extends api.HandlebarsApplicationMix
     const data = this.document.system.defenses;
 
     // Physical defenses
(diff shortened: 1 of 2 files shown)


aaclayton (approved):
Good fixes, approved.

PR #1353, opened 2026-06-23
[#1299] Address remaining `@ref[name]`s which should be `@ref[item.name]`s

roth-michael:
Now that the other bomb changes are in, can do this without conflicting with them.
This is _purely_ a replacement of `@ref[name]` to `@ref[item.name]` within the 11 actions.


aaclayton (approved):
LGTM!

PR #1357, opened 2026-06-25
[#1351, #1325] Improve ruler labeling of forced movement, implement mechanism for overriding calculated cost

### LINKED ISSUE #1325: After using Evasive Shot, you cannot make a free move
As mentioned here:
https://discord.com/channels/170995199584108546/1116696333894570094/1517272022781001941

### LINKED ISSUE #1351: Token Ruler displays costs incorrectly when Forced movement is involved
See below:
https://github.com/user-attachments/assets/8f1256b1-4098-4804-bbf6-fd71c41f4073

This is part of a more general issue with being able to appropriately rebuild movement context when looking at movement history. Related: https://github.com/foundryvtt/crucible/issues/1325 - in that if Evasive Shot is modified to _not_ provide the `hasMoved` status, it will incorrectly show as having cost 1A, despite it having been free.

roth-michael:
Closes #1351 
Closes #1325 
This is primarily two changes, one to do with forced movement, one which I think will come in handy for talents which let you move for a given static cost, like Evasive Shot.
Changes to **forced movement**:
- Any drag movement made by the GM with Forced Movement toggled on on the tools will default to the **push** movement action.
- Movement made with the **push** action will now display the distance moved on the ruler waypoints, rather than just "0ft | Forced" (set `measure` to `true` on the action config).
  - As a consequence of this change `costMultiplier: 0` is no longer sufficient to allow Infinity-cost movement (e.g. Restrained).
- Whether movement is forced is now detected based on movement action (**push** is forced, others are not) in the token ruler. This addresses the #1351 bug where whether it read Forced was entirely based on the state of the GM's tools at the time of rendering.
- Similarly, early exit from `CrucibleToken#_(pre|on)UpdateMovement` is determined by whether all passed & pending waypoints for the given movement are **push**-action.

New **measureOptions** changes:
- `CrucibleTokenObject#_getMovementCostFunction` now check for the existence of an `overrideCost` field in `options`. If present, the returned function is one which will invariably return the value of `overrideCost`.
- To support this:
  - `ActionMovementUsage` (aka `someAction.usage.movement`) now supports an optional `measureOptions`.
  - In the Action Use Dialog, these `measureOptions` are passed to `planMovement`, and to `CrucibleTokenObject#measureMovementPath`.
  - The latter function above has been temporarily (I think) overridden, as there is [maybe a core bug](https://discord.com/channels/170995199584108546/1421197270375989348/1519727582084333691) where `measureOptions` are not passed to it from `Token##updatePlannedMovement`, so if there is a planning context matching the token in question we will infer them from said context.

And an example of using the latter changes, **Evasive Shot**:
- Sets `this.usage.freeMove` in `prepare` in case an Evasive Shot was the first movement of the actor's turn.
- Sets `this.usage.movement.measureOptions` to `{overrideCost: 0}` in `prepare` to make use of the above changes and ensure that the cost for any planned movement calculates to 0 (it is still constrained by walls and by the `maxDistance`).
- Removes any potential update to `status.hasMoved` in `postActivate` to ensure that free movement isn't consumed by this _special_ free movement.

**Open Question**: Should **push** be visible on the token hud as a movement action? It seems unnecessary with the Forced Movement toggle, and like a tool that players needn't have a use for. That said I don't think it's doing much _harm_ being there.

### DIFF
--- a/module/canvas/_module.mjs
+++ b/module/canvas/_module.mjs
@@ -58,7 +58,7 @@ export function configure() {
     icon: "fa-solid fa-explosion",
     img: "icons/svg/explosion.svg",
     teleport: false,
-    measure: false,
+    measure: true,
     visualize: true,
     costMultiplier: 0,
     speedMultiplier: 1,
--- a/module/canvas/token-ruler.mjs
+++ b/module/canvas/token-ruler.mjs
@@ -42,7 +42,7 @@ export default class CrucibleTokenRuler extends foundry.canvas.placeables.tokens
       units: grid.units,
       total: (waypoint.measurement.distance - state.priorDistance).toNearest(0.01).toLocaleString(game.i18n.lang)
     });
-    const isForced = game.user.isGM && !!ui.controls.controls.tokens?.tools?.forcedMovement?.active;
+    const isForced = waypoint.action === "push";
     if ( isForced ) context.cost = {units: "", delta: game.i18n.localize("TOKEN.MOVEMENT.COST.Forced"),
       displayCost: isFinalOfSubpath};
     else if ( !Number.isFinite(deltaCost) ) context.cost = {units: "",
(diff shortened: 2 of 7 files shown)


aaclayton (approved):
Reviewed for merge with some follow-ups planned.

PR #1375, opened 2026-07-06
Automate Eye of the Storm boon

XinSysVTT:
**Main:**

Eye of the Storm gives +2 boons on a Strike/Spell against an enemy who acted during your delay. This automates it instead of leaving it as a manual reminder.

**How:** `flags.crucible.delay` already tracks your delay for the round and sticks around until your resumed turn ends. So whenever a combatant's turn starts, I check for any other actor currently delaying who considers them an enemy, and add them to a `witnessed` list on that delay flag. Then the talent's `prepareAttack` hook just checks if your target is in your own witnessed list and adds the boons if so.

No extra cleanup needed since `witnessed` lives inside `delay`, which already gets deleted when your resumed turn ends.

Tested manually: delayed, let an enemy act, struck them on my resumed turn — boons applied. Confirmed allies acting in between don't count, and the bonus goes away once the delayed turn ends.

### DIFF
--- a/_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
+++ b/_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
@@ -37,7 +37,7 @@ system:
           duration:
             value: 1
             units: rounds
-            expiry: turnEnd
+            expiry: turnStart
           system:
             dot: []
             maintenance: null
--- a/_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
+++ b/_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
@@ -37,7 +37,7 @@ system:
           duration:
             value: 1
             units: rounds
-            expiry: turnStart
+            expiry: turnEnd
           system:
             dot: []
             maintenance: null
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1048,7 +1048,7 @@ export default class CrucibleActor extends Actor {
     await this.update(foundry.utils.mergeObject(actorUpdates, {"flags.crucible.delay": {
       round: game.combat.round,
       from: combatant.initiative,
-      fromTurn: combatant.flags.crucible?.turnNumber ?? 0,
+      fromTurn: combatant.turnNumber ?? 0,
       to: initiative
     }
     }));
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1048,6 +1048,7 @@ export default class CrucibleActor extends Actor {
     await this.update(foundry.utils.mergeObject(actorUpdates, {"flags.crucible.delay": {
       round: game.combat.round,
       from: combatant.initiative,
+      fromTurn: combatant.flags.crucible?.turnNumber ?? 0,
       to: initiative
     }
     }));
--- a/module/documents/combat.mjs
+++ b/module/documents/combat.mjs
@@ -114,8 +114,33 @@ export default class CrucibleCombat extends foundry.documents.Combat {
   /** @override */
   async _onStartTurn(combatant, context) {
     await super._onStartTurn(combatant, context);
+    if ( game.user.isActiveGM ) await this.#recordDelayWitnesses(combatant);
     // TODO forward turn events to the system subtype
     return combatant.actor.onStartTurn(context);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Record enemy Actors who act while another Actor is Delaying, for talents like Eye of the Storm.
+   * @param {Combatant} combatant     The Combatant whose turn is starting
+   */
+  async #recordDelayWitnesses(combatant) {
+    const actingActor = combatant.actor;
+    if ( !actingActor ) return;
+    const round = this.round;
+    const updates = [];
+    for ( const other of this.combatants ) {
+      const delayingActor = other.actor;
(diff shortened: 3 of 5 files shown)


roth-michael (changes requested):
I don't updating potentially multiple actors each turn change is the way to go about this. When an Actor delays, `flags.crucible.delay` has a `from` of their original initiative. We can simply compare that `from` against the target's combatant's initiative at the time of attack to determine whether they took their turn between the original delay and the current turn. The only thing we would need to account for there is a combatant with the same initiative as the `from` initiative. To do that, we could potentially also include a `fromTurn` (or something) in `flags.crucible.delay` recording the combatant's original `turnNumber`. Then the comparison would be simple; if the target's combatant's `turnNumber` is greater than or equal to the attacker's actor's `flags.crucible.delay.fromTurn`, but less than the attacker's combatant's `turnNumber`, grant the boons.

It also seems that an unrelated change in Unarmed Combat Training made its way into this PR.

Also (and this doesn't matter if we go with the approach outlined above, but FYI) `Combat#_onStartTurn` only ever runs for a single GM client, so if we _do_ add anything in there, we don't need to do the `if ( game.user.isActiveGM )` check.

XinSysVTT:
Taking a second look, appreciate the guidance

XinSysVTT:
Let me know if this helps: 
Reworked Eye of the Storm's automation per review: instead of writing a "witnessed enemies" list to every delaying actor on each turn change, `Combat#_onStartTurn` now just stamps a single monotonic `turnNumber` onto the Combat and the starting Combatant (two lightweight writes, no fan-out, no GM check needed since that hook only runs on one client). `delay()` records `fromTurn` from that stamp when the actor delays, and the `prepareAttack` hook simply compares the target's and attacker's stamped turn numbers against `fromTurn` to decide whether the target acted in between — no stored list, no per-turn scanning of other combatants. Confirmed no stray edit to Unarmed Combat Training made it into the diff.

_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
```diff
@@ -37,7 +37,7 @@ system:
           duration:
             value: 1
             units: rounds
-            expiry: turnStart
+            expiry: turnEnd
```
roth-michael: Unrelated change is still here.

module/documents/actor.mjs
```diff
@@ -1048,6 +1048,7 @@ export default class CrucibleActor extends Actor {
     await this.update(foundry.utils.mergeObject(actorUpdates, {"flags.crucible.delay": {
       round: game.combat.round,
       from: combatant.initiative,
+      fromTurn: combatant.flags.crucible?.turnNumber ?? 0,
```
roth-michael:
```suggestion
      fromTurn: combatant.turnNumber ?? 0,
```
`Combatant` instances already have a `turnNumber` property.

module/documents/combat.mjs
```diff
@@ -114,6 +114,9 @@ export default class CrucibleCombat extends foundry.documents.Combat {
   /** @override */
   async _onStartTurn(combatant, context) {
     await super._onStartTurn(combatant, context);
+     const turnNumber = (this.flags.crucible?.turnNumber ?? 0) + 1;
+     await this.update({"flags.crucible.turnNumber": turnNumber});
+     await combatant.update({"flags.crucible.turnNumber": turnNumber});
```
roth-michael:
```suggestion
```
`Combatant` instances already have a `turnNumber` property.

module/hooks/talent.mjs
```diff
@@ -544,6 +544,23 @@ HOOKS.echolocation0000 = {
...
+HOOKS.eyeofthestorm000 = {
+  prepareAttack(item, action, target, rollData) {
+    if ( !action.tags.has("strike") && !action.tags.has("spell") ) return;
+    const delay = this.flags.crucible?.delay;
+    if ( !delay ) return;
+    if ( this.getDispositionTowards(target) !== CONST.TOKEN_DISPOSITIONS.HOSTILE ) return;
+    const targetTurn = target.combatant?.flags.crucible?.turnNumber;
+    const myTurn = this.combatant?.flags.crucible?.turnNumber;
+    if ( (targetTurn === undefined) || (myTurn === undefined) ) return;
+    if ( (targetTurn >= delay.fromTurn) && (targetTurn < myTurn) ) {
```
roth-michael:
```suggestion
    const targetTurn = target.combatant?.turnNumber;
    if ( !Number.isNumeric(targetTurn) ) return;
    if ( (targetTurn >= delay.fromTurn) && (targetTurn < (this.combatant?.turnNumber ?? 0)) ) {
```
Just a bit of cleanup given the existence and properties of the `turnNumber` field; that first early return will be the case if either `target.combatant` does not exist or it _does_ exist but the combat is not started. Could return early if `this.combatant?.turnNumber` is nullish too (for the same reasons), but defaulting to `0` works fine anyway, `targetTurn < 0` is effectively `false`.

roth-michael (changes requested):

XinSysVTT:
made those changes Micheal, appreciate the support, I know my coding is shit hahaha

roth-michael (approved):

aaclayton (approved):
This approach looks sensible to record the original turn number as well as the original initiative. Nice change @XinSysVTT. Thanks @roth-michael for the review help.

PR #1377, opened 2026-07-07
Added support for proposed function 861 aka opening a door in combat

### LINKED ISSUE #861: Proposal: Opening/Closing Doors cost 1 AP
Consider adding a action cost to closing and opening doors.

In Combat 
- opening and closing a door could cost 1 AP
- could intercept door handling
- could check distance to door < 2 microgrids to owned token

Advantages
- Doors can be used as tactical element, by closing those to keep enemies away/slow them down
- Nervous players cant open/close doors all the time
- Doors cann not be magically opened across the map
- Players/Characters cant open door - shoot - close door and be safe without a realistic cost

Disadvantages
- More complex interaction
- More movement fiddling
- Maybe a little costly AP

XinSysVTT:
# Main
Apologies first of all, first time doing an implementation like this, so I don't think I followed best practices. 

But in essence:

## Goal
Doors cost 1 AP to open/close in combat

### PR body:

Implements #861. Opening or closing a door in combat now costs 1 AP and requires being within 5ft of it.

New `module/hooks/door.mjs` hooks `preUpdateWall` to check distance and AP before allowing the toggle, spending AP via the existing `alterResources` helper. Only enforced in combat, and only when a token is selected. Added a `doorActionCost` world setting (default on) so GMs can disable it.

### DIFF
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -18,7 +18,6 @@ import * as models from "./module/models/_module.mjs";
 import * as audio from "./module/audio.mjs";
 import * as hooks from "./module/hooks/_module.mjs";
 import {applyEmberPatches} from "./module/hooks/ember.mjs";
-import "./module/hooks/door.mjs";
 
 // Helpers
 import {handleSocketEvent} from "./module/socket.mjs";
@@ -851,6 +850,27 @@ Hooks.on("canvasTearDown", () => {
   CONFIG.Canvas.polygonBackends.move = foundry.canvas.geometry.ClockwiseSweepPolygon;
 });
 
+
+/* -------------------------------------------- */
+
+/**
+ * Door open/close cost, refund, actor, and turn handling live entirely on the "openDoor" CrucibleAction (see
+ * module/hooks/action.mjs), which resolves the acting actor, checks turn order, and applies its Action cost
+ * atomically alongside the wall update. A normal left-click on the door icon is already routed into that Action
+ * by CrucibleDoorControl (see module/canvas/door-control.mjs), so this hook doesn't need to handle the common
+ * case - it's the backstop that stops a raw wall update from reaching the door state some other way (e.g. a
+ * macro, another module, or a stray API call) while the cost setting is active in combat. GMs are exempt.
+ */
+Hooks.on("preUpdateWall", (wall, changes, options, userId) => {
+  if ( options.crucibleAction ) return;                    // came from the openDoor action itself - let it through
+  if ( !("ds" in changes) ) return;                        // not a door-state change, ignore
+  if ( game.user.isGM ) return;                             // GMs may freely toggle doors
+  if ( !game.settings.get("crucible", "doorActionCost") ) return;
+  if ( !game.combat?.started ) return;                      // combat only
+
+  ui.notifications.warn(game.i18n.localize("WARNING.DoorUseAction"));
+  return false;
+});
 
 /* -------------------------------------------- */
 
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -18,6 +18,7 @@ import * as models from "./module/models/_module.mjs";
 import * as audio from "./module/audio.mjs";
 import * as hooks from "./module/hooks/_module.mjs";
 import {applyEmberPatches} from "./module/hooks/ember.mjs";
+import "./module/hooks/door.mjs";
 
 // Helpers
 import {handleSocketEvent} from "./module/socket.mjs";
@@ -26,7 +27,6 @@ import * as chat from "./module/chat.mjs";
 import * as interaction from "./module/interaction.mjs";
 import CrucibleTalentNode from "./module/const/talent-node.mjs";
 import {statusEffects} from "./module/const/statuses.mjs";
-
 // Party
 let party = null;
 
@@ -373,6 +373,15 @@ Hooks.once("init", async function() {
     config: false,
     type: Boolean,
     default: false
+  });
+
+  game.settings.register("crucible", "doorActionCost", {
+    name: "SETTINGS.DoorActionCostName",
+    hint: "SETTINGS.DoorActionCostHint",
+    scope: "world",
+    config: true,
+    type: Boolean,
+    default: true
   });
 
   const fields = foundry.data.fields;
@@ -841,6 +850,7 @@ Hooks.on("canvasReady", () => {
 Hooks.on("canvasTearDown", () => {
   CONFIG.Canvas.polygonBackends.move = foundry.canvas.geometry.ClockwiseSweepPolygon;
 });
+
 
 /* -------------------------------------------- */
 
--- a/lang/en.json
+++ b/lang/en.json
@@ -2065,7 +2065,9 @@
     "EnableVFXHint": "Play visual effects animations on the canvas when actions are confirmed.",
     "EnableVFXName": "Enable Action Visual Effects",
     "FlexibleAncestryAbilitiesHint": "When enabled, hero characters may reassign their Ancestry's Primary and Secondary Ability Score bonuses to abilities of their choice during character creation. The Ancestry still contributes the same +2 and +1 starting bonuses; only the abilities receiving them are flexible.",
-    "FlexibleAncestryAbilitiesName": "Flexible Ancestry Ability Bonuses"
+    "FlexibleAncestryAbilitiesName": "Flexible Ancestry Ability Bonuses",
+    "DoorActionCostName": "Doors Cost Action Points",
+    "DoorActionCostHint": "Opening or closing a door in combat costs 1 Action Point for the interacting creature."
   },
   "SKILL": {
     "CATEGORY": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -63,6 +63,10 @@
         "Description": "Move a distance by spending an amount of Action that depends on your Stride and which movement action is used.",
         "Name": "Move"
       },
+      "OpenDoor": {
+        "Description": "Open or close a door within reach. During Combat this costs 1 Action; using it again to reverse your own most recent toggle of the same door refunds that Action.",
+        "Name": "Open/Close Door"
+      },
       "ReactiveStrike": {
         "Description": "Perform a strike when an enemy leaves your engagement and you are not fully engaged.",
         "Name": "Reactive Strike"
@@ -425,6 +429,9 @@
         },
         "ESTOCADE": {
           "RequiresOpening": "Estocade requires a target who is Exposed, Prone, Slowed, Staggered, Paralyzed, or Disarmed."
+        },
+        "OPEN_DOOR": {
+          "WrongTurn": "You may only open or close a door on your own turn in combat."
         },
         "REVIVE": {
           "RequiresDead": "Revive requires a Dead target."
@@ -2694,6 +2701,7 @@
     "CannotEquipInvalidCategory": "Actor {actor} cannot use {item} in the {type} equipment slot.",
     "CannotEquipSlotInUse": "Actor {actor} cannot equip {item} because the {type} item slot is already in use.",
     "CannotEquipTaxonomy": "Actor {actor} cannot equip {item} because the \"{taxonomy}\" taxonomy is incapable of using equipment.",
+    "DoorUseAction": "Use the Open/Close Door action from your sheet to interact with doors during combat.",
     "NoParty": "You must have a primary party configured for this.",
     "SkillRequireAncestryBackground": "You must assign both Ancestry and Background before you can purchase Skill ranks."
   },
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -1488,6 +1488,24 @@ export const DEFAULT_ACTIONS = Object.freeze([
     autoFavorite: true
   },
 
+  // Open/Close Door
+  {
+    id: "openDoor",
+    name: "ACTION.DEFAULT_ACTIONS.OpenDoor.Name",
+    img: "icons/environment/wilderness/cave-entrance.webp",
+    description: "ACTION.DEFAULT_ACTIONS.OpenDoor.Description",
+    target: {
+      type: "self",
+      number: 0,
+      scope: 1
+    },
+    cost: {
+      action: 1
+    },
+    tags: ["harmless"],
+    autoFavorite: true
+  },
+
   // Reactive Strike
   {
     id: "reactiveStrike",
--- a/module/hooks/action.mjs
+++ b/module/hooks/action.mjs
@@ -1,3 +1,5 @@
+import {requestDoorToggle} from "../socket.mjs";
+
 const HOOKS = {};
 
 /* -------------------------------------------- */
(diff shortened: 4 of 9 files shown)


lang/en.json
```diff
@@ -2065,7 +2065,9 @@
     "EnableVFXHint": "Play visual effects animations on the canvas when actions are confirmed.",
     "EnableVFXName": "Enable Action Visual Effects",
     "FlexibleAncestryAbilitiesHint": "When enabled, hero characters may reassign their Ancestry's Primary and Secondary Ability Score bonuses to abilities of their choice during character creation. The Ancestry still contributes the same +2 and +1 starting bonuses; only the abilities receiving them are flexible.",
-    "FlexibleAncestryAbilitiesName": "Flexible Ancestry Ability Bonuses"
+    "FlexibleAncestryAbilitiesName": "Flexible Ancestry Ability Bonuses",
+    "DoorActionCostName": "Doors Cost Action Points",
+    "DoorActionCostHint": "Opening or closing a door in combat costs 1 Action Point for the interacting creature."
```
roth-michael, line 2077: Very minor thing, but try to keep i18n keys alphabetical (all `NAMESPACES` alphabetically first, then all `ActualKeys` alphabetically).

crucible.mjs
```diff
@@ -18,6 +18,7 @@ import * as models from "./module/models/_module.mjs";
 import * as audio from "./module/audio.mjs";
 import * as hooks from "./module/hooks/_module.mjs";
 import {applyEmberPatches} from "./module/hooks/ember.mjs";
+import "./module/hooks/door.mjs";
```
roth-michael: Right now, the `module/hooks` folder has a very specific meaning; it's not hooks in the "normal" foundry sense, but explicitly action/actor hooks registered based on embedded items on an actor. So I don't _think_ this is where we'd want to add the new logic, if indeed we do approach it this way. Right now all our "normal" hooks are just registered in `crucible.mjs` directly. I wouldn't make any changes yet though, see my main review comment.

crucible.mjs
```diff
@@ -842,6 +851,7 @@ Hooks.on("canvasTearDown", () => {
   CONFIG.Canvas.polygonBackends.move = foundry.canvas.geometry.ClockwiseSweepPolygon;
 });
 
+
```
roth-michael, line 853: Sneaky extra newline I'm only mentioning as something to look out for.

module/hooks/door.mjs
```diff
@@ -0,0 +1,30 @@
...
+  const token = canvas.tokens.controlled[0];
+  if ( !token?.actor ) return;                            // nobody selected, don't enforce
+  const actor = token.actor;
+
+  // Distance check
+  const wallMidpoint = {x: (wall.c[0] + wall.c[2]) / 2, y: (wall.c[1] + wall.c[3]) / 2};
+  const {distance} = canvas.grid.measurePath([token.center, wallMidpoint]);
+  const MAX_DOOR_RANGE = 5; // distance from the door
+  if ( distance > MAX_DOOR_RANGE ) {
+    ui.notifications.warn(`${actor.name} is too far from the door to interact with it.`);
```
roth-michael, line 18: User-facing warnings like this we'd want to localize.

module/hooks/door.mjs
```diff
@@ -0,0 +1,30 @@
...
+    return false;
+  }
+
+  // Action point check
+  if ( actor.resources.action.value < 1 ) {
+    ui.notifications.warn(`${actor.name} does not have enough Action Points to open/close this door.`);
```
roth-michael, line 24: User-facing warnings like this we'd want to localize.

module/hooks/door.mjs
```diff
@@ -0,0 +1,30 @@
...
+  if ( !game.combat?.started ) return;                    // combat only
+
+  const token = canvas.tokens.controlled[0];
+  if ( !token?.actor ) return;                            // nobody selected, don't enforce
+  const actor = token.actor;
+
+  // Distance check
+  const wallMidpoint = {x: (wall.c[0] + wall.c[2]) / 2, y: (wall.c[1] + wall.c[3]) / 2};
+  const {distance} = canvas.grid.measurePath([token.center, wallMidpoint]);
+  const MAX_DOOR_RANGE = 5; // distance from the door
```
roth-michael, line 16: We _might_ want this to exist somewhere more concrete rather than be ephemerally created. In fact, a general "interaction range" might make sense as a property on each actor. Not suggesting you make this change, just musing.

roth-michael (commented):
I think, and I'm not the guy making final decisions here, but I _think_ we'd probably want to handle this at least a little bit differently.

The change I'm almost certain we'd want to make is having some "open door" synthetic action be used, which in turn could be reversed to close the door and refund the Action. Naturally it would be an imperfect reverse, as fog of war isn't something we can easily reverse.

What I'm not as sure about is whether we'd want it to happen as a result of clicking the door, or whether we'd want to do something interesting like disable interaction with doors for players with a token selected, and _create_ an action on the sheet (or maybe un-hide a latent one) in their favorites when their token is within range of a door which could be used directly from the sheet.

In any case, I'd wait and see what @aaclayton has to say about it before investing any more time into this.

aaclayton (changes requested):
Thanks for the review comments @roth-michael, agreed with your callouts. The most significant problems are in `preUpdateWall` with the wrong actor and lacking combat/turn/combatant awareness. Watch out for non-atomic transactions with possible race conditions.

XinSysVTT:
Lookin into these changes will get something for this

XinSysVTT:
### `Add door-toggle Action economy (openDoor) with GM-arbitrated race safety`

Adds a new `openDoor` Action so an actor can open/close the nearest door within reach, either from their sheet or by clicking the door icon directly on the canvas. In combat this costs 1 Action (gated by the new `doorActionCost` world setting) and enforces turn order. Toggling the same door again to undo your last flip refunds the Action automatically through the existing event-reversal pipeline. Outside combat, or with the setting off, it's free and unrestricted.

A couple of things worth flagging:

- Originally you could only trigger this from the sheet button — clicking the door icon itself just hit a wall (pun intended), since raw wall updates get blocked by `preUpdateWall` while the cost setting is active. Wasn't sure which UX we wanted so I made both work: the sheet button and a direct door-icon click now both route through the same Action.
- While working on nearest-door detection I realized a door on a level above/below you could technically end up "closer" in raw distance than one on your own floor in some layouts. Added a floor check to rule those out — let me know if that's overkill.
- Toggling wasn't previously safe against two people clicking near-simultaneously (each reads the door state independently, both writes land, last one wins). Fixed by having non-GM toggles go through the active GM's client via socket request/response, with a per-wall queue on the GM side so concurrent requests for the same door serialize and each one reads live state right before acting. Unsure if this is best practice or not so let me know! 

Open to feedback on any of this, especially the door-icon-click piece since that wasn't explicitly asked for.

XinSysVTT (commented):
Please let me know any recommendations / changes

lang/en.json
```diff
@@ -63,6 +63,10 @@
         "Description": "Move a distance by spending an amount of Action that depends on your Stride and which movement action is used.",
         "Name": "Move"
       },
+      "OpenDoor": {
+        "Description": "Open or close a door within reach. During Combat this costs 1 Action; using it again to reverse your own most recent toggle of the same door refunds that Action.",
```
roth-michael, line 67: Description here is a bit verbose. It can be assumed what reversing the action will do, and the cost will already be provided via tag wherever this shows up.

lang/en.json
```diff
@@ -426,6 +430,9 @@
         "ESTOCADE": {
           "RequiresOpening": "Estocade requires a target who is Exposed, Prone, Slowed, Staggered, Paralyzed, or Disarmed."
         },
+        "OPEN_DOOR": {
+          "WrongTurn": "You may only open or close a door on your own turn in combat."
+        },
```
roth-michael, line 435: I think we'll want a more generic localization, something like the opposite of Reaction that's just "You may only use the {action} action on your own turn in combat."

module/const/action.mjs
```diff
@@ -1489,6 +1488,24 @@ export const DEFAULT_ACTIONS = Object.freeze([
...
+      number: 0,
+      scope: 1
+    },
+    cost: {
+      action: 1
+    },
+    tags: ["harmless"],
+    autoFavorite: true
+  },
+
```
roth-michael, line 1508: I think it would make the most sense not to have a specific "Open Door" action, but rather an "Interact" or "Manipulate" action.

This action we'd both suppress from the sheet as you're already doing in `hooks/action.mjs` but _also_ not auto-favorite, for now at least. Currently the only valid use for it would be opening doors, and I think it's cleaner if the action is prompted only by clicking a door on the scene explicitly, rather than dealing with "I want to open _a door_ so I'll click it from my sheet" and figuring out which door they mean.

This'd be extensible in the future to apply to other potential environmental interactions. In ember, for instance, we could wire up the interactable stuff so that they cost some action while in combat. Then we would be able to reuse this generic action. 

A further thought that just occurred to me: Down the line we could potentially show this on the sheet again, but have it automatically change its name/cost depending on what is providing its availability. Definitely not in scope for this PR, but just more reason to start with a more generic action.

module/hooks/action.mjs
```diff
@@ -1538,6 +1549,110 @@ HOOKS.omniglotDecoction = {
...
+    if ( forcedId !== undefined ) delete action.usage.forcedDoorWallId;
+    if ( forcedId ) {
+      const wall = canvas.walls?.get(forcedId);
+      // The explicitly targeted door may no longer exist or be out of reach - don't silently substitute whatever
+      // door happens to be nearest, since that isn't the door the player clicked.
+      if ( wall?.document.isDoor && HOOKS.openDoor._isReachable(action.actor, wall) ) return wall;
+      return null;
+    }
+    return HOOKS.openDoor._nearestDoor(action.actor);
+  },
```
roth-michael, line 1624: If we do go with the "this action can't be used directly from the sheet" suggestion, most of this can be removed as we make the assumption that there is an explicit wall ID passed, for which we simply need to do the distance & other `canUse` checks.

module/hooks/door.mjs
```diff
@@ -0,0 +1,30 @@
+/* -------------------------------------------- */
+/*  Door Open cost Function                       */
+/* -------------------------------------------- */
+Hooks.on("preUpdateWall", (wall, changes, options, userId) => {
```
roth-michael, line 4: This file is now vestigial.

module/models/action.mjs
```diff
@@ -1327,6 +1338,14 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
       }
     }
 
+    // Seed any one-shot usage overrides onto this bound Action *before* cloning. #clone passes this.usage to the
+    // clone's construction context by reference (see #clone), and the clone re-runs prepare() synchronously as
+    // part of that construction - so anything merged in here is visible to the Action's own prepare() hook before
+    // it would otherwise compute a default (e.g. HOOKS.openDoor guessing the nearest door). Individual hooks are
+    // responsible for consuming and clearing any one-shot keys they read, so a value passed for this single use()
+    // call doesn't linger on the shared usage object and leak into a later, unrelated preparation.
+    if ( usage ) foundry.utils.mergeObject(this.usage, usage, {inplace: true});
```
roth-michael, line 1347: I think we can avoid this (& the other related changes to this file). Elsewhere, when specific `usage` is needed (`forcedTargets` being used, for instance) the action is simply created, then usage set, _then_ `use` called. I see no reason not to do the same here. This would mean either grabbing, cloning, amending, & using the action rather than calling `CrucibleActor#useAction` in `CrucibleDoorControl` _or_ augmenting `CrucibleActor#useAction` to check for `usage` and if present perform the necessary logic itself.

crucible.mjs
```diff
@@ -26,7 +26,6 @@ import * as chat from "./module/chat.mjs";
 import * as interaction from "./module/interaction.mjs";
 import CrucibleTalentNode from "./module/const/talent-node.mjs";
 import {statusEffects} from "./module/const/statuses.mjs";
-
```
roth-michael: What happened to this newline?

crucible.mjs
```diff
@@ -375,6 +374,15 @@ Hooks.once("init", async function() {
...
 
+  game.settings.register("crucible", "doorActionCost", {
+    name: "SETTINGS.DoorActionCostName",
+    hint: "SETTINGS.DoorActionCostHint",
+    scope: "world",
+    config: true,
+    type: Boolean,
+    default: true
+  });
+
```
roth-michael, line 385: I don't think we actually need to make this a setting. During combat, opening doors costs Action. If someone doesn't like that, the GM can always step in and open the door themselves.

crucible.mjs
```diff
@@ -842,6 +850,28 @@ Hooks.on("canvasTearDown", () => {
...
+  if ( options.crucibleAction ) return;                    // came from the openDoor action itself - let it through
+  if ( !("ds" in changes) ) return;                        // not a door-state change, ignore
+  if ( game.user.isGM ) return;                             // GMs may freely toggle doors
+  if ( !game.settings.get("crucible", "doorActionCost") ) return;
+  if ( !game.combat?.started ) return;                      // combat only
+
+  ui.notifications.warn(game.i18n.localize("WARNING.DoorUseAction"));
+  return false;
+});
+
```
roth-michael, line 874: I'm torn on whether to include this. On the one hand, there's something to be said for enforcing this rule. On the other, I think it would probably be cleaner if we just ensured that _we_ never break convention in our own codebase, and don't get in the way of what modules or macros might do. As long as we handle it appropriately in-system, the main thing this does is provide a potential stumbling block for someone writing a system agnostic module.

module/socket.mjs
```diff
@@ -32,3 +38,111 @@ function _onReplayActionVFX({messageId}) {
   action.playVFXEffect(vfxConfig, references);
 }
 
+/* -------------------------------------------- */
```
roth-michael, line 41: I know Andrew said "Watch out for non-atomic transactions with possible race conditions" but I think this infrastructure is overkill. We effectively have two scenarios to consider, the "door change via action" scenario and the "door change not via action" scenario. The latter scenario is what happens when a GM clicks a door, or if we opt to skip the action workflow when out of combat rather than just make the action free. That scenario is also what every other system on foundry that doesn't explicitly code for door state changes gets for free, because core handles users clicking doors by itself.

So really we only need to look at the "door change via action" scenario. I'd suggest that the thing to do here would just be to store the door's intended or current state (either works, they're just inverses) somewhere persistent on the action itself. For the sake of example I'll say we store the intended state, but if we wanted to store the current (time of use) state the logic would be extremely similar. Then, in the `confirm` hook, check the door's current state. Compare wall's state at time of hook call with action-intended wall state. If `reverse` and the states match, update the wall so that they don't. If _not_ `reverse` and the states _don't_ match, update the wall so that they do. Do nothing otherwise.

Concrete example:
1. Door is CLOSED, it is combat.
2. Click door.
3. Action 1 workflow - since door is CLOSED, intended state is OPEN, store on action/message.
4. Confirm Action 1. Current state is CLOSED, intended state is OPEN, update wall document to OPEN.
5. Action 2 workflow - since door is OPEN, intended state is CLOSED, store on action/message.
6. Reverse Action 1. Current state is OPEN, intended state is OPEN, update wall document to CLOSED.
7. Confirm Action 2. Current state is CLOSED, intended state is CLOSED, do not update wall document.
8. Reverse Action 2. Current state is CLOSED, intended state is CLOSED, update wall document to OPEN.

As you can (hopefully, not sure how clear that was) see, Action 1 will always say "I want this door to open when confirmed, and close when reversed," and Action 2 will always say "I want this door to close when confirmed, and open when reversed." Making actions aware of the intended state means that confirming or reversing a given action will _always_ result in the appropriate wall state for that specific action being confirmed/reversed, rather than an action just toggling the current state.

roth-michael (changes requested):
Some more feedback here. The main callout I think would be getting rid of the "Open Door" action in favor of a programmatically-used-only "Interact" action, but there's a good amount of suggestions throughout the comments.

Issue #751, opened 2026-03-18
Address certain effects _not_ having their duration based on the applied-to actor's combatant


roth-michael:
See: https://discord.com/channels/170995199584108546/1116696333894570094/1482525422649610374

Primary example is Grappled. When grappling somebody, on success the source gets "Grappling" and the target gets "Grappled" - "Grappled" and "Grappling" should have the exact same lifespan, but we currently make the assumption that all applied effects base their durations on their parent. 

The other example in that message, Formidable Presence, is probably just a matter of modifying expiry to be end of turn, or modifying the duration to be 2 rounds and updating the description; I see no reason to have it adhere to different duration rules the way Grappled needs to.

If all "need different rules" examples happen to be a parent & child relationship, then I think we could make the child have an indefinite duration, and ensure interdependency between the two effects (if one is deleted, so is the other).

aaclayton:
Feels like this isn't easily fixable as a "bugfix" and we need some more architecture to handle relationship between seemingly unrelated effects.

PR #1382, opened 2026-07-09
Adding support for grappling ideas - v3

### LINKED ISSUE #1323: Issue with the Grappled effect
“Grappled” has a grapple effect that ends at the beginning of the turn. Shouldn't it end at the end of the turn?

Image

Additionally, Grappled does not have a removal DC.

Image

XinSysVTT:
### Problem

The Grapple action creates two paired effects — `Grappling` on the attacker and `Grappled` on the target — linked only by `origin`. In practice this pairing wasn't actually enforced outside of a few hardcoded spots:

- Ending one side of a grapple (deleting the effect from a sheet/HUD, natural expiry, the grappler or target dying, or an actor being deleted outright) did **not** reliably end the other side. Only movement-breaking, a successful Escape check, and Throw explicitly handled both sides.
- Both effects were authored with a **1-round duration** (`expiry: turnStart` / `turnEnd`), so grapples silently ended at a turn boundary regardless of whether anyone tried to escape.
- Neither effect had a `dc` set, so it defaulted to `null` → `Infinity` on data prep. This made the Grappled effect **invisible to the Escape action** (`HOOKS.escape._restraints()` only considers effects with a finite `dc`), and there was no way to say "auto-compute this DC" for an effect authored directly on an Item — `dc: null` already means "unremovable," leaving no third state for "figure it out."
- The Action effect editor template (`effect.hbs`) never exposed a Removal DC field at all, so there was no UI path to set one on an action-authored effect like Grapple's.

### Changes

**`module/documents/active-effect.mjs`**
Added a general-purpose pairing cleanup in `_onDelete()`:
```js
await this.#endGrapplePair();
```
`#endGrapplePair()` checks if the deleted effect is the Grappling or Grappled effect, resolves the paired actor via `origin`, and deletes their matching effect too — regardless of *how* this effect was removed (manual delete, natural expiry, etc). Existing hardcoded breaks (movement, Escape, Throw) are unaffected; they just become redundant no-ops once the paired effect is already gone.

**`module/documents/actor.mjs`**
Added `#endGrappleOnDeath()`, called from `#applyResourceStatuses()` when an actor's health update pushes them into `isDead`. Deletes this actor's own Grappling/Grappled effect (if any); the pairing cleanup above cascades to the counterpart. A dead actor can no longer maintain or meaningfully resist a grapple.

**`crucible.mjs`**
Added a `deleteActor` hook. Deleting an Actor document outright doesn't run the normal deletion workflow for its embedded ActiveEffects, so the `_onDelete` pairing cleanup never fires for the counterpart in that case. This hook handles it explicitly by checking the deleted actor's Grappling/Grappled effect and cleaning up the partner directly.

**`module/models/effect-base.mjs`**
Added an `AUTO_DC` sentinel (`-1`) and handled it in `prepareBaseData()`:
```js
static AUTO_DC = -1;

prepareBaseData() {
  if ( this.dc === null ) this.dc = Infinity;
  else if ( this.dc === CrucibleBaseActiveEffect.AUTO_DC ) {
    const actor = this.parent?.parent;
    this.dc = SYSTEM.PASSIVE_BASE + (actor?.system.advancement?.level ?? 0);
  }
}
```
This generalizes DC auto-fill (previously only handled inline in `Action#finalizeEvents()` for effects created through the action/event pipeline) to **any** effect, including ones authored directly on an Item — since `null` already means "unremovable," authors now set `dc: -1` to request "auto-compute" instead. Computes as `12 (PASSIVE_BASE) + the effect bearer's own level`.

**`templates/sheets/action/effect.hbs`**
Added the missing Removal DC field to the Action effect editor, bound to `system.fields.dc` (which already existed in the schema, just wasn't rendered anywhere for action-authored effects):
```hbs
{{formGroup fields.system.fields.dc name=(concat effect.fieldPath ".system.dc") value=effect.system.dc
            classes="slim" placeholder="∞"}}
```

**`_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml`** (and recompiled `packs/talent`)
Updated the Grapple action's Grappled and Grappling effects:
- Removed the 1-round `duration` on both (now `duration: {}` — persistent until something actually ends it).
- Set `dc: -1` on both, so each computes a real removal/escape DC (`12 + bearer's level`) instead of `Infinity`.

### Known limitation / follow-up

The auto-computed DC scales with the **bearer's own level** for both effects — meaning the DC to escape a grapple currently scales with the *victim's* level, not the grappler's Athletics skill. A weak and a highly skilled grappler produce the same escape DC against the same target. If opposed-skill-based DC is wanted instead, that would mean explicitly setting `dc` in `HOOKS.grapple.postActivate()` (which already sets `grappling.origin`) rather than relying on the generic `AUTO_DC` fallback. Not addressed in this PR.

### Testing notes

- Grapple actor A → actor B; confirm both `Grappling`/`Grappled` effects persist indefinitely (no turn-boundary expiry).
- Delete the `Grappling` effect manually (sheet or HUD) → confirm `Grappled` on the target is also removed, and vice versa.
- Reduce the grappler (or target) to 0 HP → confirm the grapple ends on both sides.
- Delete the grappler (or target) Actor outright → confirm the counterpart's effect is cleaned up.
- Inspect `Grappled.system.dc` on a freshly-created effect → should be a finite number (`12 + level`), not `Infinity`.
- Note: effects created *before* this fix was compiled retain their old snapshot data (1-round duration, `dc: null`) — existing grants of the Unarmed Combat Training talent need to be removed and re-added to pick up the corrected Grapple action data.

# PS: 

This was a complete test case, please suggest changes!

### DIFF
--- a/_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
+++ b/_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
@@ -34,11 +34,9 @@ system:
             all: false
           statuses:
             - restrained
-          duration:
-            value: 1
-            units: rounds
-            expiry: turnStart
+          duration: {}
           system:
+            dc: -1
             dot: []
             maintenance: null
             regions: []
@@ -51,11 +49,9 @@ system:
             all: false
           statuses:
             - restrained
-          duration:
-            value: 1
-            units: rounds
-            expiry: turnStart
+          duration: {}
           system:
+            dc: -1
             dot: []
             maintenance: null
             regions: []
--- a/crucible.mjs
+++ b/crucible.mjs
@@ -806,6 +806,25 @@ Hooks.on("getSceneControlButtons", controls => {
 Hooks.on("renderCombatTracker", models.CrucibleCombatChallenge.onRenderCombatTracker);
 
 /* -------------------------------------------- */
+/*  grappling solution                          */
+/* -------------------------------------------- */
+Hooks.on("deleteActor", async actor => {
+  if ( !game.user.isActiveGM ) return;
+  const grapple = crucible.api.hooks.action.grapple;
+  const pairs = {
+    [grapple._GRAPPLED_EFFECT_ID]: grapple._GRAPPLING_EFFECT_ID,
+    [grapple._GRAPPLING_EFFECT_ID]: grapple._GRAPPLED_EFFECT_ID
+  };
+  for ( const [ownId, partnerId] of Object.entries(pairs) ) {
+    const effect = actor.effects.get(ownId);
+    if ( !effect?.origin ) continue;
+    const partner = await fromUuid(effect.origin);
+    const partnerEffect = partner?.effects?.get(partnerId);
+    if ( partnerEffect ) await partnerEffect.delete();
+  }
+ });
+
+/* -------------------------------------------- */
 /*  Canvas Hooks                                */
 /* -------------------------------------------- */
 
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -80,6 +80,21 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
         if ( token ) await token.delete();
       }
     }
+    await this.#endGrapplePair();
+  }
+
+
+  async #endGrapplePair() {
+    const grapple = crucible.api?.hooks?.action?.grapple;
+    if ( !grapple ) return;
+    const partnerId = {
+      [grapple._GRAPPLED_EFFECT_ID]: grapple._GRAPPLING_EFFECT_ID,
+      [grapple._GRAPPLING_EFFECT_ID]: grapple._GRAPPLED_EFFECT_ID
+    }[this.id];
+    if ( !partnerId || !this.origin ) return;
+    const partner = await fromUuid(this.origin);
+    const partnerEffect = partner?.effects?.get(partnerId);
+    if ( partnerEffect ) await partnerEffect.delete();
   }
 
   /* -------------------------------------------- */
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -2965,6 +2965,15 @@ export default class CrucibleActor extends Actor {
     }
   }
 
+
+  async #endGrappleOnDeath() {
+    const grapple = crucible.api?.hooks?.action?.grapple;
+    if ( !grapple ) return;
+    const ids = [grapple._GRAPPLED_EFFECT_ID, grapple._GRAPPLING_EFFECT_ID]
+    .filter(id => this.effects.has(id));
+    if ( ids.length ) await this.deleteEmbeddedDocuments("ActiveEffect", ids);
+  }
+
   /* -------------------------------------------- */
 
   /**
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -2958,6 +2958,8 @@ export default class CrucibleActor extends Actor {
       await this.toggleStatusEffect("weakened", {active: this.system.isWeakened && !this.system.isDead });
       await this.toggleStatusEffect("dead", {active: this.system.isDead});
       await this.toggleStatusEffect("asleep", {active: false});
+      if ( this.system.isWeakened || this.system.isDead ) await this.#endGrappleOnDeath();
+
     }
     if ( ("morale" in r) || ("madness" in r) ) {
       await this.toggleStatusEffect("broken", {active: this.system.isBroken && !this.system.isInsane });
--- a/module/models/effect-base.mjs
+++ b/module/models/effect-base.mjs
@@ -2,7 +2,7 @@
  * Active Effect subtype containing crucible-specific system schema.
  */
 export default class CrucibleBaseActiveEffect extends foundry.data.ActiveEffectTypeDataModel {
-
+static AUTO_DC = -1;
   /** @inheritDoc */
   static defineSchema() {
     const fields = foundry.data.fields;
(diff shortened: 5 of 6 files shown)


XinSysVTT:
this is for #1323

module/documents/active-effect.mjs
```diff
@@ -80,6 +80,21 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
...
+    const grapple = crucible.api?.hooks?.action?.grapple;
+    if ( !grapple ) return;
+    const partnerId = {
+      [grapple._GRAPPLED_EFFECT_ID]: grapple._GRAPPLING_EFFECT_ID,
+      [grapple._GRAPPLING_EFFECT_ID]: grapple._GRAPPLED_EFFECT_ID
+    }[this.id];
+    if ( !partnerId || !this.origin ) return;
+    const partner = await fromUuid(this.origin);
+    const partnerEffect = partner?.effects?.get(partnerId);
+    if ( partnerEffect ) await partnerEffect.delete();
```
roth-michael, line 97: This makes sense, but I wonder whether we won't have additional cases for mutually dependent effects int he future and, if so, whether we might not prefer a more generic handling for this. Definitely an Andrew question.

module/documents/actor.mjs
```diff
@@ -2958,13 +2958,24 @@ export default class CrucibleActor extends Actor {
       await this.toggleStatusEffect("weakened", {active: this.system.isWeakened && !this.system.isDead });
       await this.toggleStatusEffect("dead", {active: this.system.isDead});
       await this.toggleStatusEffect("asleep", {active: false});
+      if ( this.system.isWeakened || this.system.isDead ) await this.#endGrappleOnDeath();
```
roth-michael, line 2961: Weakened shouldn't necessarily be grounds for ending a grapple, I don't think. Especially for one being grappled.

But moreover, similar to my comment about paired effects, I think if we do indeed want to handle this (rather than having this be a "GM deletes certain effects when the actor dies" thing), we'd want a more generic solution where Active Effects themselves define whether they should be removed upon parent actor death.

module/documents/actor.mjs
```diff
@@ -2958,13 +2958,24 @@ export default class CrucibleActor extends Actor {
...
     if ( ("morale" in r) || ("madness" in r) ) {
       await this.toggleStatusEffect("broken", {active: this.system.isBroken && !this.system.isInsane });
       await this.toggleStatusEffect("insane", {active: this.system.isInsane});
     }
   }
 
+
+  async #endGrappleOnDeath() {
+    const grapple = crucible.api?.hooks?.action?.grapple;
+    if ( !grapple ) return;
```
roth-michael, line 2973: This applies to `#endGrapplePair` as well, but I think if we're in a world where `grapple` doesn't exist here, something has gone very very wrong. We could nix the optional chaining and the existence check.

module/models/effect-base.mjs
```diff
@@ -2,7 +2,7 @@
  * Active Effect subtype containing crucible-specific system schema.
  */
 export default class CrucibleBaseActiveEffect extends foundry.data.ActiveEffectTypeDataModel {
-
+static AUTO_DC = -1;
```
roth-michael, line 5:
```suggestion
  static AUTO_DC = -1;
```

module/models/effect-base.mjs
```diff
@@ -32,8 +32,11 @@ export default class CrucibleBaseActiveEffect extends foundry.data.ActiveEffectT
 
   /** @override */
   prepareBaseData() {
-    // An effect with no removal DC cannot be removed; represent that as an infinite difficulty
     if ( this.dc === null ) this.dc = Infinity;
+    else if ( this.dc === CrucibleBaseActiveEffect.AUTO_DC ) {
+        const actor = this.parent?.parent;
+        this.dc = SYSTEM.PASSIVE_BASE + (actor?.system.advancement?.level ?? 0);
+      }
```
roth-michael, line 39: You mention it indirectly in your write-up, but this is too early. An effect should be allowed to exist with a `-1` DC, and if it has one it should be the job of the Action which creates it to override that value.

crucible.mjs
```diff
@@ -805,6 +805,25 @@ Hooks.on("getSceneControlButtons", controls => {
...
+  };
+  for ( const [ownId, partnerId] of Object.entries(pairs) ) {
+    const effect = actor.effects.get(ownId);
+    if ( !effect?.origin ) continue;
+    const partner = await fromUuid(effect.origin);
+    const partnerEffect = partner?.effects?.get(partnerId);
+    if ( partnerEffect ) await partnerEffect.delete();
+  }
+ });
+
```
roth-michael, line 826: I don't think we need to track this. Actor deletions should be rare, and them happening during a grapple involving said actor even more rare.

_source/talent/Unarmed_Combat_Training_unarmedCombatTra.yml
```diff
@@ -51,11 +49,9 @@ system:
             all: false
           statuses:
             - restrained
-          duration:
-            value: 1
-            units: rounds
-            expiry: turnStart
+          duration: {}
```
roth-michael, line 52: I think we may still want the grappler's effect to expire at the start of their next turn. Not 100% sure.

roth-michael (commented):
Somewhat similar to my thoughts on opening doors in combat; I think we may want to make a more generic solution here. It would, I think, be an easier generic solution to make this time though. Again, would hold off on making additional changes until @aaclayton has weighed in.

aaclayton (changes requested):
If we want to head in this direction, I think we need a more generalized solution for effect pairing that isn't exclusive to grappling.  This requires some significant system design thought that probably is not the right thing to track in this PR. My recommendation is to record this experiment as a github issue for posterity (there isn't one currently) that we can resurface when a generalized underlying foundation exists.

roth-michael:
Related: #751 which _somewhat_ covers this.

PR #1384, opened 2026-07-09
Fix enchantment attack bonus when enchantment tier is derived from affixes

### LINKED ISSUE #1378: Enchantment Tier set by affixes does not add Enchantment Bonus to attack rolls
If you add affixes to a weapon it sets and locks the enchantment tier of the weapon correctly, but if you attack with the weapon it does not add the enchantment bonus to the dice roll:

Image

Image

However, editing the enchantment tier of a weapon manually correctly adds the enchantment bonus:

Image

Image

### LINKED ISSUE #1396: Armour's dodge does not increase if affixes added before modifying enchantment tier
As title says, if an affix is added before the enchantment tier is manually modified, the armour's dodge defense won't increase. This is the same issue weapons had with attack not increasing (https://github.com/foundryvtt/crucible/issues/1378) but i did not see armours being mentioned in that issue's pull request.

Take any armour, modify the enchantment tier first then add an affix, the dodge will increase as intended.
Image

Now take the same armour, add the affix directly and let the enchantment tier be modified by the affixes, and the dodge will not increase.
Image

XinSysVTT:
# Main

Fixes #1378.

### Problem

`actionBonuses.enchantment` was being initialized during `prepareBaseData()`, before an item's enchantment tier had been recalculated from its affixes.

For weapons whose enchantment tier is provided by affixes:

* the displayed enchantment tier was correct,
* the tier was correctly locked,
* but attack rolls continued using the pre-derived enchantment value (typically `mundane`), resulting in no enchantment bonus.

Manually setting the enchantment tier worked because that value already existed during `prepareBaseData()`.

### Cause

Affix-derived enchantment is computed in `CruciblePhysicalItem.prepareDerivedData()`, after Active Effects have been initialized. `Weapon.prepareBaseData()` was caching the enchantment bonus before that override occurred.

### Fix

* Initialize `actionBonuses.enchantment` to `0` in `prepareBaseData()`.
* Populate it in `prepareDerivedData()`, after `super.prepareDerivedData()` has applied the affix-derived enchantment tier.

This ensures attack bonuses always reflect the final computed enchantment tier, whether it comes from a manually assigned value or from affixes.

### DIFF
--- a/module/models/item-armor.mjs
+++ b/module/models/item-armor.mjs
@@ -58,7 +58,7 @@ export default class CrucibleArmorItem extends CruciblePhysicalItem {
    */
   prepareBaseData() {
     super.prepareBaseData();
-    const {category, quality, enchantment} = this.config;
+    const {category, quality} = this.config;
 
     // Armor Defense
     this.armor.base = Math.clamp(this.armor.base, category.armor.min, category.armor.max);
@@ -66,7 +66,7 @@ export default class CrucibleArmorItem extends CruciblePhysicalItem {
 
     // Dodge Defense
     this.dodge ||= {};
-    this.dodge.base = category.dodge.base(this.armor.base) + enchantment.bonus;
+    this.dodge.base = category.dodge.base(this.armor.base);
     this.dodge.scaling = category.dodge.scaling;
 
     // Natural applies property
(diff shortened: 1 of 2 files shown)


aaclayton (changes requested):
Looks correct, but `item-armor.mjs` has the same bug which is unfixed in this PR. Need to fix both locations.

XinSysVTT:
Added a second commit that should address the armor issues #1396

XinSysVTT (commented):
Fix for both item armor and item weapon

aaclayton (approved):

PR #1386, opened 2026-07-12
Fix: Create & Conjure summons now replace the previous creature instead of orphaning it

### LINKED ISSUE #1301: Create & Conjure gestures do not adhere to "only one at a time" rules
And because of this, they don't gracefully handle multiple summons (because it shouldn't be possible).

XinSysVTT:
## Fix: Create & Conjure summons now replace the previous creature instead of orphaning it

potential fix for  #1301

### Problem
Create and Conjure track their summon on an `ActiveEffect` with a deterministic id, so recasting is meant to just refresh that effect. But `summon.confirm()` always created a new token first, so the "refresh" was really an update, not a delete+create — `_onDelete` never fired, and the old token was left orphaned on the scene instead of being cleaned up.

### Fix
In `summon.confirm()`, before creating the new token, delete whatever token the existing same-id effect is currently tracking:

```js
const priorEffectIds = new Set(summonEvents.map(e => e.summon.effectId).filter(Boolean));
for ( const effectId of priorEffectIds ) {
  const existing = this.actor.effects.get(effectId);
  if ( !existing ) continue;
  for ( const uuid of existing.system.summons ?? [] ) {
    const token = await fromUuid(uuid);
    if ( token ) await token.delete();
  }
}
```

### Testing
Cast Conjure twice — only one creature should remain. Same for Create. Cast Conjure then Create — both should coexist (different gestures, different effect ids) -- unsure if this is how we want this so please let me know!

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -615,6 +615,17 @@ export const TAGS = {
       const summonEvents = this.events.filter(e => e.type === "summon");
       if ( !summonEvents.length ) return;
 
+       // Enforce "only one at a time": dismiss whatever creature this gesture's deterministic effect id is currently tracking, before summoning a new one.
+      const priorEffectIds = new Set(summonEvents.map(e => e.summon.effectId).filter(Boolean));
+       for ( const effectId of priorEffectIds ) {
+         const existing = this.actor.effects.get(effectId);
+         if ( !existing ) continue;
+         for ( const uuid of existing.system.summons ?? [] ) {
+           const token = await fromUuid(uuid);
+           if ( token ) await token.delete();
+         }
+       }
+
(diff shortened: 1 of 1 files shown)


aaclayton (changes requested):
This one took me a while to review because as I pulled on the thread it kept leading one level deeper to a more fundamental root problem. 

Where this led me, and what I think we ultimately need is https://github.com/foundryvtt/crucible/pull/1397 which addresses the problem by deleting (with a newly hardened permission boundary) previous tokens or regions that were formerly associated with the effect that have been removed via an update operation.

aaclayton:
Resolved via https://github.com/foundryvtt/crucible/pull/1397

PR #1387, opened 2026-07-12
Fix Alchemist's Fire effect description reference

### LINKED ISSUE #1310: Description ref name bug in the Alchemist's Fire burning effect.
Being hit by the Alchemist's Fire flask correctly applies the Burning condition, but the effect's description text replaces "Alchemist's Fire" with "Burning"

Image

XinSysVTT:
### Summary

Fixes an issue where the **Burning** effect applied by **Alchemist's Fire** displays an incorrect description.

Previously, when the Burning condition was applied, the description would incorrectly replace references to **"Alchemist's Fire"** with **"Burning"**, resulting in misleading text.

### Root Cause

The effect description was being enriched after the applied effect's name had been auto-set, causing `@ref` resolution to use the effect name (`Burning`) instead of the originating action (`Alchemist's Fire`).

### Changes

* Preserve the original action context when resolving `@ref` references for effect descriptions.
* Resolve `@ref` references during action use rather than after the effect has been auto-generated.
* Leave other enrichers to be processed lazily/as needed to avoid changing existing enrichment behavior.

### Result

The Burning effect now correctly references **Alchemist's Fire** in its description while preserving the existing enrichment pipeline for other description content.

### Fixes

potential fix for #1310

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -1816,17 +1816,8 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
 
   /* -------------------------------------------- */
 
-  /**
-   * Resolve only @ref[...] links in a description against this action, leaving every other enricher (@Rule,
-   * @Condition, etc.) untouched so they still enrich normally whenever the description is later rendered.
-   * Must run before this action's description is copied onto a created effect (see #recordEffectEvents), since
-   * gesture/item hooks (e.g. Alchemist's Fire) can overwrite that effect's own data - including its name - with
-   * generic status data (e.g. "Burning") before the effect is ever displayed. Resolving @refs now, while `this`
-   * still has valid item/gesture context, keeps references like "@ref[item.name]" pinned to the source item's
-   * name instead of silently drifting to whatever the effect's data happens to hold later. See GH #1310.
-   * @param {string} text
-   * @returns {string}
-   */
+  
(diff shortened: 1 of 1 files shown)


module/models/action.mjs
```diff
@@ -1816,6 +1816,18 @@ export default class CrucibleAction extends foundry.abstract.DataModel {
...
 
+  
+ /** Resolve @ref[...] links against this action before copying the description to an effect, preserving source references if the effect's data changes later. */
+  #resolveRefEnrichers(text) {
+    if ( !text ) return text;
+    return text.replace(/@ref\[([\w.]+)](?:{([^}]+)})?/g, (match, path, fallback) => {
+      const attr = foundry.utils.getProperty(this, path);
+      return attr ?? fallback ?? match;
+    });
+  }
```
aaclayton, line 1824: Instead of reinventing the regex and replacement logic (which could possibly drift), we should route this through the registered enricher handler, for example:
```js
#resolveRefEnrichers(text) {
  const {pattern, enricher} = CONFIG.TextEditor.enrichers.find(e => e.id === "reference");
  return text.replace(pattern, (...args) => enricher(args.slice(0, -2), {relativeTo: this}).textContent);
}
```

aaclayton (changes requested):
Good plan, but we should handle `#resolveRefEnrichers` in a smarter way

XinSysVTT (commented):
adding the fix recommended

aaclayton (approved):
Good idea, thank you.

PR #1392, opened 2026-07-16
Refactor attack resolution to trigger hooks for all outcomes, includi…

### LINKED ISSUE #1152: Call `receiveAttack` for misses in addition to glances & hits
Use Case: **Swashbuckler**, which _automatically_ reduces morale of the attacker whose attack was dodged or parried.

XinSysVTT:
## Call `receiveAttack` on all results; implement Swashbuckler

**Problem:** `weaponAttack`/`spellAttack` only fired the `receiveAttack` actor hook on Glance/Hit, so defensive talents couldn't react to a dodged/parried attack. Separately, the Swashbuckler talent had rules text but no code behind it.

**Changes:**
- `module/documents/actor.mjs` (+ mirrored in `crucible-compiled.mjs`): removed the `result >= GLANCE` guard on `receiveAttack` in `weaponAttack` and `spellAttack`, so the hook now fires for every result, matching `skillAttack` and the Interposer.
- `module/hooks/talent.mjs` (+ compiled): added `HOOKS.swashbuckler0000.receiveAttack`, draining the attacker's morale (equal to the Swashbuckler's Presence) on Dodge/Parry.

**Not included:** Swashbuckler's crit → nearby-allies-gain-morale effect (no existing passive aura pattern to follow; happy to add in a follow-up).

**For awarness:** Existing `receiveAttack` hooks (`armoredInstinct0`, `swarm`, `duelist`) already guard on specific result types and no-op safely on a miss -- I dont think this would impact but please let me know if you think this could should differ

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -845,7 +845,7 @@ export default class CrucibleActor extends Actor {
     // Create and evaluate the AttackRoll instance, then resolve its outcome and structured damage
     const roll = new AttackRoll(rollData);
     await roll.evaluate();
-    const result = roll.resolveDamage(this, target, {
+    roll.resolveDamage(this, target, {
       multiplier: rollData.multiplier,
       base: spell.damage.base,
       bonus: rollData.damageBonus + (this.system.rollBonuses.damage?.[rollData.damageType] ?? 0),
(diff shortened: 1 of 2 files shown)


aaclayton (approved):
Looks exactly right. Great change!

PR #1393, opened 2026-07-16
Refactor hazard label preparation for improved clarity and efficiency

### LINKED ISSUE #1385: Render hazard enrichers in a more terse format with tags that apply to the hazard reserved only for the expanded tooltip.
For example in:

Image

It's fine for this to just read "Boulderfall (Hazard 12)" as long as the tooltip explains the Reflex defense and Bludgeoning damage tags.

XinSysVTT:
fix for : [#1385](https://github.com/foundryvtt/crucible/issues/1385)

## Terser hazard enrichers

**Change:** `module/enrichers.mjs` (+ compiled bundle) — `enrichHazard` no longer lists every tag's label inline. `[[/hazard 12 reflex bludgeoning]]{Boulderfall}` now renders as `Boulderfall (Hazard 12)` instead of `Boulderfall (Hazard 12, Reflex, Bludgeoning)`. Defense/damage/resource detail already exists in the hover tooltip, untouched.

**Caveat:** tags outside defense/damage/resource (e.g. `severe`) previously showed inline; now dropped entirely, not moved to tooltip. Flagging as unsure if we want this or not

**Test:** paste `[[/hazard 12 reflex bludgeoning]]{Boulderfall}` into a journal page, check inline text + hover tooltip.

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -506,14 +506,32 @@ function enrichHazard([_match, terms, name]) {
 
   // Prepare label
   const hazardRank = _loc("HAZARD.Rank", {danger});
-  const label = name ? _loc("HAZARD.Parenthetical", {label: name, tags: hazardRank}) : hazardRank;
+  const parenthetical = name ? [hazardRank] : [];
+  const severeCfg = SYSTEM.ACTION.TAGS.severe;
+  if ( tags.includes("severe") && severeCfg?.label ) parenthetical.push(_loc(severeCfg.label));
+  let label = name || hazardRank;
+  const listFormatter = new Intl.ListFormat(game.i18n.lang, {style: "short", type: "unit"});
+  if ( parenthetical.length ) label = _loc("HAZARD.Parenthetical", {label, tags: listFormatter.format(parenthetical)});
+
   // Prepare tooltip
   const defenseLabel = _loc(SYSTEM.DEFENSES[defenseType]?.label);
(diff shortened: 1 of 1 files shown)


roth-michael:
Part of me thinks we should make a special exception for `severe` and allow that inline. That one feels particularly relevant. Either way, I think we should definitely move all tags into the tooltip; there needs to be _some_ way to show exactly what the configured hazard is without clicking it.

aaclayton (changes requested):
1. I agree with @roth-michael we should make a special exception for `severe` in particular that should always remain inline, i.e. `Boulderfall (Hazard 12 Severe)`
2. I don't think we should silently drop any tag. If the tag doesn't have an informative tooltip to display for the tag we should at least display its name without further information somehow.
3. For UI changes like this, please always include a screenshot of what it looks like after the change (particularly the tooltip)

XinSysVTT:
Added a new commit with this fix: 
{0ABF6243-FB88-4400-819A-BD948C3B8139}

XinSysVTT (commented):
New full fix if it works

XinSysVTT:
{9B55834A-44CD-436B-A9DE-72395FAB0214}

{2EB02227-36EC-41B8-B49A-EE3B37E39472}

aaclayton (approved):
Good change. I want to do a follow-up change to try and improve the localization, but I'll handle that as a follow-up amendment.

PR #1397, opened 2026-07-16
Implement a hardened approach for active effect updates which remove or replace managed Region or Token documents

aaclayton:
While working on https://github.com/foundryvtt/crucible/pull/1386: as I pulled on the thread it kept leading one level deeper to a more fundamental root problem. 

Where this led me, and what I think we ultimately need is https://github.com/foundryvtt/crucible/pull/1397 which addresses the problem by deleting (with a newly hardened permission boundary) previous tokens or regions that were formerly associated with the effect that have been removed via an update operation.

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -10,6 +10,19 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
    */
   static TOOLTIP_TEMPLATE = "systems/crucible/templates/tooltips/tooltip-active-effect.hbs";
 
+  /**
+   * Document types the owned-reference deletion cascade is permitted to delete.
+   * @type {Set<string>}
+   */
+  static #DELETABLE_TYPES = new Set(["Token", "Region"]);
+
+  /**
+   * The UUIDs of the Tokens and Regions this effect owns. Cached from the effect's own persisted data.
+   * The responsible active GM derives the set of documents to delete from trusted, replicated state.
+   * @type {Set<string>}
+   */
+  #ownedReferences;
+
   /* -------------------------------------------- */
 
   /**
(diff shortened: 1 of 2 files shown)


aaclayton:
@XinSysVTT Take a look at this branch and see if you feel that it addresses the root issue #1301 as an alternative approach to #1386

module/documents/active-effect.mjs
```diff
@@ -65,20 +78,42 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
...
+   * @returns {Set<string>}
+   */
+  #collectOwnedReferences() {
+    return new Set([...(this.system.summons ?? []), ...(this.system.regions ?? [])]);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Delete owned references on behalf of the active GM, enforcing that each still exists, is an allowed document type,
```
roth-michael: Minor thing here, but it's not on behalf of the active GM; rather, the active GM is deleting on behalf of the user performing the update/deletion.

module/documents/active-effect.mjs
```diff
@@ -164,6 +199,19 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
...
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _onUpdate(changed, options, userId) {
+    await super._onUpdate(changed, options, userId);
+
+    // Delete any owned Token/Region no longer referenced after this update
+    const prior = this.#ownedReferences ?? new Set();
+    this.#ownedReferences = this.#collectOwnedReferences();
+    const removed = [...prior].filter(uuid => !this.#ownedReferences.has(uuid));
```
roth-michael:
```suggestion
    const removed = prior.difference(this.#ownedReferences);
```
Might as well have some fun with sets.

module/documents/active-effect.mjs
```diff
@@ -228,6 +276,14 @@ export default class CrucibleActiveEffect extends foundry.documents.ActiveEffect
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _onDelete(options, userId) {
+    await super._onDelete(options, userId);
+    await this.#deleteOwnedReferences(this.#collectOwnedReferences(), userId); // Delete every owned Token/Region
```
roth-michael: Is there any reason we re-collect here rather than just passing `this.#ownedReferences`?

roth-michael (approved):
LGTM, no significant suggestions.

Wondering if we might soon extend this to handle the "X effect depends on Y effect" use case.

aaclayton:
Thanks @roth-michael good callouts.

PR #1401, opened 2026-07-18
Implement invisible enemy handling in flanking calculations and updat…

### LINKED ISSUE #1391: Invisible creatures count for flanking, which reveals their position to players
Invisible creatures count for flanking, which reveals their position to players

Source: Ember Team Playtest

XinSysVTT:
fix #1391 

based on the chat on discord I went with the idea of: 
- you cannot benefit from flanked if you cannot see the ally flanking to help, thus flanked does not get applied instead the invisable attacker gains a +1 boon. let me know y'all thoughts. 

### Problem
Invisible creatures currently contribute to flanking (as flankers, and as allies
reducing a target's flanked stage). Since a target's flanked status/boon is visible
mechanical information, this let players infer "something invisible is here" purely
from watching flanking numbers change — an unintended meta.

### Fix
- **`module/canvas/token.mjs`** — `computeFlanking()` now skips any creature with the
  `invisible` status when counting flankers *and* when computing the ally-bonus
  reduction. Invisibility no longer shows up in flanking math at all, on either side.
- **`module/documents/actor.mjs`** — `_configureAttackerRollData()` grants an invisible
  attacker a personal `+1` boon on **melee attacks only** (mirrors the existing
  flanked-boon's `isRanged` restriction). The boon is flagged `hidden: true`.
- **`templates/dice/partials/standard-check-details.hbs`** — hidden boons still render
  into the chat card (so the roll math/tooltip is correct) but get a `gm-only` CSS
  class instead of a label-only line.
- **`module/documents/chat-message.mjs`** — `onRenderHTML` strips `.gm-only` elements
  for non-GM viewers (same pattern already used to hide summoned-creature names from
  players), so the GM sees the boon's source and players just see the roll total.

### Notes
- Scope is intentionally limited to the `invisible` status; other stealth-like
  statuses are untouched.

### Testing
- Manually verified in Foundry: an invisible enemy adjacent to a target no longer
  raises `flanked`; an invisible ally no longer reduces it.
- Verified the invisible attacker's melee attack boon appears in the GM's chat log
  breakdown but not in a player's, while the roll total is identical for both.

### DIFF
--- a/module/canvas/token.mjs
+++ b/module/canvas/token.mjs
@@ -602,9 +602,11 @@ export default class CrucibleTokenObject extends foundry.canvas.placeables.Token
     engagement.allyBonus = 0;
 
     // Count flankers; an adversary's flankingStrength lets it count as more than one
+    // Invisible enemies do not contribute to flanking - their advantage is a personal, secret boon instead
     let flankers = 0;
     for ( const enemy of engagement.enemies ) {
       const {isBroken, isIncapacitated} = enemy.actor.system;
+      if ( enemy.actor.statuses.has("invisible") ) continue;
       if ( !(isBroken || isIncapacitated) ) flankers += enemy.actor.system.movement?.flankingStrength ?? 1;
     }
     engagement.flankers = flankers;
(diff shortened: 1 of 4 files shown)


XinSysVTT:
also was wondering if invisable would you gain a boon no matter the distance eg range attack or not, was unsure so I kept it close to the original and as on discord as possible

aaclayton:
I'm exploring a solution to this via https://github.com/foundryvtt/crucible/issues/1424

aaclayton:
Closing in favor of https://github.com/foundryvtt/crucible/pull/1425

PR #1408, opened 2026-07-20
[#1402] Don't fully gate setting `targetLabel` behind GM check

### LINKED ISSUE #1402: Attacks which target normal defenses (rather than passive skills for instance) display "DC" in chat card rather than "Physical"/"Reflex" etc. for non-GMs

roth-michael:
Closes #1402


aaclayton (approved):
Good fix, thanks!

PR #1410, opened 2026-07-20
[#1406] Only render Private Biography for actor owners

### LINKED ISSUE #1406: "Private Biography" field is visible to Users with only `limited` permissions, despite the tooltip suggesting it should be visible only to those with `owner` permissions

roth-michael:
Closes #1406


aaclayton (approved):
I want a small change to gate in `#prepareBiography` so we don't spend time enriching private text for non-owners, but I can do that cleanup post-merge.

PR #1413, opened 2026-07-21
[#1412] Check all knowledge (not only background) for `hasKnowledge`, and allow Adversaries to have knowledge areas

### LINKED ISSUE #1412: Manually-added knowledge is treated as nonexistent for passive knowledge enricher tooltips
Turns out we never made `CrucibleActor#hasKnowledge` look at the _actual knowledge_, only the background-granted stuff.

roth-michael:
Closes #1412 
Technically, the only thing that's needed to close #1412 is the change in `CrucibleActor#hasKnowledge` - however, when fixing this I realized that currently:
1. The tooltips under Knowledge Areas & Known Languages for Adversaries when they don't have any incorrectly assume a Hero actor sheet is being viewed.
2. The configure button for Knowledge Areas throws an error for Adversaries if pressed.

So I decided hey, why not creep the scope a little bit:
1. Added new simpler variants of those tooltips, and reference them in an if/else in the hbs itself.
2. Add `details.knowledge` to Adversary's schema, move Hero-sheet-only `knowledge` context prep up into base actor sheet. No special data prep is required for the adversary subtype itself, since nothing currently grants them knowledge areas.

If this is a change you don't want to make, I can re-submit with just the `hasKnowledge` fix.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -881,8 +881,10 @@
       "FavoriteActions": "Favorite Actions",
       "Knowledge": "Knowledge Areas",
       "KnowledgeHint": "Learn some Knowledge Areas from your Background or edit your Hero to gain some.",
+      "KnowledgeHintAdversary": "No Knowledge Areas known.",
       "Languages": "Known Languages",
       "LanguagesHint": "Learn some Languages from your Background or edit your Hero to gain some.",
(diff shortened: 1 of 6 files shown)


aaclayton (approved):
Looks right, thanks!

PR #1415, opened 2026-07-26
Bug Fix: interpose hook to handle asynchronous damage resolution

XinSysVTT:
## Fix: Interpose reaction not applying damage to interposer

**Bug:** When a character used Interpose, the chat card showed the redirected strike's damage correctly, but the interposer's health was never actually deducted.

### Original: 
https://github.com/user-attachments/assets/5e7377b2-465e-49b6-af7f-3b87b570be91

**Root cause:** `_rewrite()` posted the rewritten strike's chat message pre-flagged as `confirmed: true`. This is meant to signal "already resolved," but it actually skipped the only code path that applies resource deltas to the actor (`ChatMessage#_onCreate` → `#autoConfirmMessage` bails out early if the message is already marked confirmed, so `CrucibleAction#confirm` → `#applyEvents` never ran). Additionally, `_rewriteStrike` never computed a resource delta from the recomputed damage in the first place.

**Fix:**
- `_rewriteStrike` is now `async` and calls a new `_resolveDamageResources` helper to convert the re-evaluated damage into an actual resource delta (mirroring how the core event pipeline does it), attached to the recorded strike event.
- `_rewrite` now creates the rewritten message unconfirmed and explicitly calls `CrucibleAction.confirmMessage()` on it, so `#applyEvents` runs and the interposer's health actually updates.

###: Fixed: 

https://github.com/user-attachments/assets/fee2b8c6-ceec-40eb-83df-4fef043ef9c1

**Testing:** Confirmed a hit-and-interpose sequence now debits the interposer's health by the recomputed damage amount instead of leaving it unchanged.

### DIFF
--- a/module/hooks/action.mjs
+++ b/module/hooks/action.mjs
@@ -1371,7 +1371,7 @@ HOOKS.interpose = {
     for ( const event of originalAction.events ) {
       if ( (event.type === "strike") && event.roll ) {
         const roll = HOOKS.interpose._cloneRoll(event.roll);
-        HOOKS.interpose._rewriteStrike(rewrittenAction, roll, event.weapon, interposer);
+        await HOOKS.interpose._rewriteStrike(rewrittenAction, roll, event.weapon, interposer);
       }
       else if ( (event.type === "activation") || (event.type === "actorUpdate") ) {
         rewrittenAction.recordEvent({type: event.type, resources: event.resources,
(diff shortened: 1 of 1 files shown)


aaclayton (approved):
Good general change, but I'm going to do some cleanup after because I think we need to use our core `#resolveEventStream` method which should become `@internal` to permit this sort of usage.

PR #1419, opened 2026-08-01
Automate Dawn Beacon (and Engulfing Darkness) light sources, darkness counteraction, and fix target acquisition

### LINKED ISSUE #1418: Dawn Beacon does not create a light source (automation required)
Dawn Beacon's description promises a pillar of light (30ft bright / 60ft dim, 6 rounds), but casting it applies no light source — it must currently be placed manually by the GM.

**Dawn Beacon** (Illumination + Create + Pull)

- Creates a pillar of magical light centered on the caster's position at the moment of casting
- Radius: 30ft bright light + additional 30ft dim light (60ft total) - not automated
- Persists for 6 rounds, or until dispelled, at the fixed location (doesn't follow the caster)  
- Counteracts and destroys any magical darkness whose source lies within its area - not automated
- Enemies within the area when the light first appears must resist a Fortitude attack or be Blinded for one round - not automated

**Expected:** An AmbientLight matching the spell's description should be created automatically at the cast location and removed after 6 rounds or on dispel.

https://github.com/user-attachments/assets/8dc6ed63-484d-477d-bce4-cd44c88d7943

Will create an PR for this issue very shortly

XinSysVTT:
fix for #1418 

**Summary:**
Dawn Beacon's light source, its darkness-counteraction clause, and its enemy targeting were all non-functional/WIP. This PR implements the missing mechanics end-to-end.

**Changes:**

- **Owned light references** (`module/models/effect-base.mjs`, `module/documents/active-effect.mjs`) — added a `lights` field (Set of AmbientLight UUIDs) to the effect schema, alongside existing `regions`/`summons`. Extended `#DELETABLE_TYPES` and owned-reference collection/cleanup to include `AmbientLight`, so a light tied to an effect is cascade-deleted when that effect expires or is dispelled.

- **Dawn Beacon** (`module/hooks/spell.mjs`, `_source/spell/Dawn_Beacon_dawnBeacon000000.yml`)
  - Added a self-scoped "Dawn Beacon Pillar" effect (6 rounds) that owns the light.
  - New `confirmAction` hook (keyed `dawnBeacon000000`, matching the item's actual `_id`) creates the pillar's `AmbientLight` (30ft bright / 60ft dim) at the caster's token on cast.
  - Before placing its own light, scans the scene for "magical darkness" sources (any `AmbientLight` with `config.negative: true`) within its 60ft radius and destroys them via their owning effect.
  - Fixed `target.scope` from `1` (Self) to `3` (Enemies) — the stored value meant `#getTargetDispositions()` returned no valid dispositions at all, so enemies were never acquired as targets (no roll prompt, no manual-target override possible). This was the root cause of blind not applying.
  - Removed WIP note, fixed duplicate `iconicSpell` tag.
  - On reverse/undo, explicitly deletes the light via the actor's live effect document rather than relying on cascade-delete (a reversed action can be a reconstructed instance without the light's UUID).

- **Engulfing Darkness** (`module/hooks/spell.mjs`, `_source/spell/Engulfing_Darkness_engulfingDarknes.yml`) — automated identically (own `AmbientLight` with `negative: true`, owned by a new "Engulfing Darkness Sphere" self effect, reverse cleanup), since it had the same WIP gap and was needed as a real target for Dawn Beacon's darkness-destruction to counteract.

**Known follow-up (not in this PR):** Engulfing Darkness still has `target.scope: 1` (same bug as Dawn Beacon had) and no effect that actually blinds creatures inside it, will make a seperate issue & PR for this

### DIFF
--- a/_source/spell/Dawn_Beacon_dawnBeacon000000.yml
+++ b/_source/spell/Dawn_Beacon_dawnBeacon000000.yml
@@ -19,7 +19,8 @@ system:
         counteracts magical darkness and destroys any source of magical darkness
         whose origin is contained within its area.</p><p>Enemies within the area
         when the light is first created must resist an attack against their
-        <strong>Fortitude</strong> or be @Condition[blinded] for one round.</p>
+        <strong>Fortitude</strong> or be @Condition[blinded] for one
+        round.</p>
       cost:
         action: 0
         focus: 0
@@ -63,7 +64,7 @@ system:
             units: rounds
             expiry: turnEnd
           result:
-            type: success
+            type: any
             all: false
           system:
             changes: []
@@ -72,6 +73,16 @@ system:
             regions: []
             summons: []
             lights: []
+            light:
+              bright: 30
+              dim: 60
+              color: "#fff2c2"
+              alpha: 0.35
+              negative: false
+              animation:
+                type: pulse
+                speed: 1
+                intensity: 2
       tags:
         - spell
         - iconicSpell
--- a/_source/spell/Dawn_Beacon_dawnBeacon000000.yml
+++ b/_source/spell/Dawn_Beacon_dawnBeacon000000.yml
@@ -19,11 +19,7 @@ system:
         counteracts magical darkness and destroys any source of magical darkness
         whose origin is contained within its area.</p><p>Enemies within the area
         when the light is first created must resist an attack against their
-        <strong>Fortitude</strong> or be @Condition[blinded] for one
-        round.</p><p><strong>WIP: This spell does not yet automatically create a
-        light source, which must be done manually for now. Eventually this will
-        be automated using a new Foundry VTT version 14 framework for template
-        placement.</strong></p>
+        <strong>Fortitude</strong> or be @Condition[blinded] for one round.</p>
       cost:
         action: 0
         focus: 0
@@ -37,7 +33,7 @@ system:
       target:
         type: pulse
         number: 1
-        scope: 1
+        scope: 3
         self: false
         size: 60
       effects:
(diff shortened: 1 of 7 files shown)


XinSysVTT:
Example of fix applied: 

https://github.com/user-attachments/assets/0682b2f1-eb52-4376-a0ae-2d69ea9b44f5

XinSysVTT:
One thing to bring up would we want this to be a one time effect? As at the moment of casting, anyone who walks into the sphere afterward won't be affected.

aaclayton (changes requested):
Good attempt @XinSysVTT but we don't have the right mechanism here for a robust long-term solution.

The shape of what this is trying to do is to have an ActiveEffect that is produced via a CrucibleAction own and maintain the presence of an AmbientLight for its duration. This is a reasonable desire and something we should do by having the crucible AE model track `lights` as it does with `regions`, `summons`, etc...

This solves a few things in a more standardized way - the AE knows that it owns the light. Removing the AE removes the light. 

No special code needed anymore, configuring an action to conjure a light source (unattached to a token) involves just configuring its self-target active effect. No bespoke reverse path (get that for free).

If you feel like you want to take on that revision, I would like to see it split up into multiple steps:

1. Changes to the ActiveEffect crucible subtype schema so that light sources can *theoretically* exist as part of an AE.

2. Changes to these spells to take advantage of that new mechanism.

XinSysVTT:
working on this, implemented a bug that i'm trying to fix, exploring the fix

XinSysVTT:
okay think I've fixed all the issues, let me know

Issue #1135, opened 2026-06-02
Consider enhancing various tooltips with more detailed information


roth-michael:
The scope of this could range from specific to fields like defense, to more generic fields such as each ability score.
Definitely in scope, if desired: Instead of just showing:
>12 + ((Strength + Wisdom) / 4)

for the Fortitude defense tooltip, show something like:
> Base = (Strength + Wisdom ) / 4 = 2.5, rounds to 3
> Bonuses = 1
> Total = 12 + Base + Bonuses

And when hovering a skill value, instead of just:
> [(Strength + Dexterity) / 4] + Skill Bonus + Enchantment Bonus

show
> Ability Bonus = (Strength + Dexterity / 4) = 1.25, rounds to 1
> Skill Bonus = -4 (Untrained)
> Enchantment Bonus = 0
> Total = Ability Bonus + Skill Bonus + Enchantment Bonus

If we wanted to go crazy, could also add explanatory flavor text when hovering over things like Strength itself.

PR #1420, opened 2026-08-10
Enhancing various tooltips with more detailed information

Streyder:
### Summary

First time I am working with curcible and foundry in general to be honest. Hope this up to standards.
Open to all feedback. 

First part of addressing  #1135. If this is acceptable, I will update other tooltips as well.
So far this only changes the Skills tooltip.

### New tooltips: 
image
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1522,7 +1522,7 @@
   },
   "DEFENSES": {
     "TOOLTIPS": {
-      "Fortitude": "{base} + ((Strength + Wisdom) / 4)",
+      "Fortitude": "{base} + [(Strength + Wisdom) / 4]",
       "Madness": "{base} + (Madness / 10)",
       "Reflex": "{base} + ((Dexterity + Intellect) / 4)",
       "Willpower": "{base} + ((Toughness + Presence) / 4)",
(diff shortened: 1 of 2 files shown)


aaclayton (approved):
Thanks for your contribution @Streyder - this is a nice first addition. 

I do want to simplify it down a little though, I don't really like having multi-line tooltips here, but I do think your suggestion of adding the actual numeric values into the tooltip is the key innovation here and that's what we should keep. 

I'm going to merge this and then apply some changes on top of it that are a little more terse in terms of the final tooltip behavior.

Streyder:
@aaclayton thanks for the feedback!

I completely agree with you, that the multi-line tooltip is a little odd here.
I also completely agree, that the key innovation here is adding the actual numeric values.

Another goal I tried to achieve was to explain how the final value is being calculated.
While players can certainly look this up in the system rules, especially because the system is so automated, I found this a nice way explain how things come to be here. I think it would be a shame to drop this.

Sadly I couldnt make both goals happen at the same time, while also being concise enough to put everything into a singular line, especially on smaller screens.

Looking forward to what your solution will look like so I can implement it for all the other tooltips as well.

PR #1421, opened 2026-08-10
Change damage type of noxious spray to poison

Streyder:
### Summary

Changes the damage type of the noxious spray attack from electricity to poison.
This is how it is described in the text.


aaclayton (approved):
Thanks for the fix!

PR #1425, opened 2026-08-16
Flanking refactor

### LINKED ISSUE #1311: Flanking overlay got stuck a few times: toggling "Visualise Flanking" off does nothing and the overlay doesn't update with token movement.
Persisted even with no active combat and no token selected (reported independently by three playtesters - refresh fixed I think)

Image

Image

### LINKED ISSUE #1391: Invisible creatures count for flanking, which reveals their position to players

### LINKED ISSUE #1424: Redesign the way that Flanking works. Compute token adjacency eagerly on movement as we currently do, but defer computation of flanking bonuses until part of the action workflows that require them.
1. Flanking is now based on the attacker's point of view that is potentially asymmetric based on perception and detection. You only benefit from allies you can perceive. 

2. Invisible creatures no longer give themselves away. An invisible creature can flank a target when attacking themselves, even if they don't contribute towards flanking for creatures who can't see them.

3. Added a new **Overrun** condition lets you impose one forced stage of Flanking directly. A GM can apply it to represent a circumstance the map doesn't capture. Overrun is more than just a bonus, because it can be used to open up a target for actions that can only be used against flanked targets like Backstab, Duelist parry, or denial of Reactive Strikes.

4. No more "Flanked" status effect or status icon, this is replaced by "Overrun".

5. The flanking visualization tool is improved to work with a combination of your CONTROLLED token and your TARGET tokens. It shows the attackers/allies/engagement and resulting flanked stage for each target from the perspective of a controlled token. It respects what that token can (or cannot) perceive.

6. A small Duelist change, which now benefits from +2 parry passively, but potentially loses that bonus if they are flanked.

7. Charging into melee now counts towards flanking. Actions with the "move" target type will treat the attacker as an eligible participant in the flanking computation regardless of their current position.

8. Fixed a couple related bugs due to flanking with Backstab and Pack Hunter.

aaclayton:
Closes #1424 
Closes #1391
Closes #1311


module/canvas/token.mjs
```diff
@@ -572,54 +602,72 @@ export default class CrucibleTokenObject extends foundry.canvas.placeables.Token
...
+   * @param {object} [options]
+   * @param {boolean} [options.includeSelf]     Count this Token as a flanker even where it is not yet engaged, as
+   *                                            when it will close to melee before it strikes
+   * @returns {CrucibleTokenFlanking}
+   */
+  getFlankingAgainst(target, {includeSelf=false}={}) {
+    if ( this === target ) return {flankers: 0, allyBonus: 0, flanked: 0};
+    const engagement = target.engagement;
+    let {enemies} = engagement;
+    if ( includeSelf && !enemies.has(this) ) enemies = enemies.union(new Set([this]));
```
roth-michael, line 672:
```suggestion
    if ( includeSelf && !enemies.has(this) ) enemies.add(this);
```

module/hooks/action.mjs
```diff
@@ -1643,30 +1643,36 @@ HOOKS.rallyingTonic = {
...
+      HOOKS.reactiveStrike._reject.call(this, "overrun");
     }
   },
+  preActivate() {
+    if ( HOOKS.reactiveStrike._championBypass(this.actor) ) return;
+    const token = this.token?.object;
+    if ( !token ) return;
+    const {flanked} = crucible.api.canvas.CrucibleTokenObject.computeFlanking(token.engagement, {observer: token});
+    if ( flanked ) HOOKS.reactiveStrike._reject.call(this, "flanked");
+  },
```
roth-michael: I think ideally we'd move this into `acquireTargets` so that the flanked issue is surfaced prior to actually clicking "Use Action"

module/models/_module.mjs
```diff
@@ -9,7 +9,6 @@ export {default as CrucibleSpellAction} from "./spell-action.mjs";
 // ActiveEffect
 export {default as CrucibleAffixActiveEffect} from "./effect-affix.mjs";
 export {default as CrucibleBaseActiveEffect} from "./effect-base.mjs";
-export {default as CrucibleFlankedActiveEffect} from "./effect-flanked.mjs";
```
roth-michael, line 12: Should remove from `documentTypes.ActiveEffect` in `system.json`

crucible.mjs
```diff
@@ -1140,6 +1102,94 @@ Hooks.once("diceSoNiceReady", dice3d => {
...
+/* -------------------------------------------- */
+
+/**
+ * Delete the ActiveEffects which the retired automatic flanking system committed to Actors, wherever they persist.
+ * Flanking is now derived per-attacker at the moment of use, so these persisted effects are inert automation
+ * artifacts. They are identified by their "flanked" subtype, which no manually applied condition ever carries.
+ * @returns {Promise<void>}
+ */
+async function _deleteFlankedEffects() {
+  console.groupCollapsed("Crucible | Retired Flanking Effect Cleanup");
```
roth-michael, line 1152: Not _really_ important, but wondering whether we can't gate this behind an explicit version check so it only ever happens once. I guess that runs the risk of newly-added old compendia not having it done, so probably not.

roth-michael (changes requested):
Looks good to me generally, now time to actually test it. One definite issue is that a token with its vision disabled, despite us constructing an ephemeral vision source, does not seem to detect any surrounding tokens. I believe this is because its `detectionModes` will be empty for a sight-disabled token due to core's `TokenDocument#_prepareDetectionModes` doing an early exit if `!this.sight.enabled` and therefore not setting `lightPerception` or `basicSight` defaults.

roth-michael:
Additional bug found via testing: Flanking should re-compute when choosing a different weapon in action use dialog (pistol mainhand, scimitar offhand, switching to scimitar will not add flanked boons in dialog or actual roll)

module/canvas/token.mjs
aaclayton: Reads awkwardly, but is intentional to avoid mutation of `enemies` which is the shared cached set of the target's engagement graph which would contaminate flanking for other (actor,target) pairs.

module/models/_module.mjs
aaclayton: Intentionally not done to allow for a migration period without causing pre-existing active effect documents to go invalid.

crucible.mjs
aaclayton: Yeah,. good idea to improve for later, but out of scope for this PR. Should adopt something closer to how Ember does its migration framework at some point.

module/canvas/token.mjs
roth-michael: Ahh, right. Makes sense.

module/hooks/action.mjs
aaclayton: Intentional to move it here so its not in `canUse` which would run `computeFlanking` on every sheet render since reactive strike is (probably) a favorited action.

crucible.mjs
roth-michael: I suppose once we remove the subtype from the manifest, we could just remove this code as well.

module/hooks/action.mjs
roth-michael: Not suggesting we use `canUse`, but the new(ish) `acquireTargets`

PR #1448, opened 2026-08-31
[#1444] Add Blood Sense as a detection mode

### LINKED ISSUE #1444: Convert "Blood Sense" to be a proper detection mode which would allow it to detect invisible creatures, work while blinded, and participate in flanking.
Since flanking now uses the tokens detected by the attacker (or rather, the tokens that _would be_ detected by the attacker if their vision was enabled), being blinded makes it impossible for a creature to benefit from flanking unless they have a detection mode that somehow bypass that.

I'd propose that we enhance Blood Sense and create a corresponding detection mode (showing non-max-health enemies with a dark red glow perhaps) which we assign similar to echolocation or thermal vision.

roth-michael:
Closes #1444

Pretty simple one. Only things worth calling out:
- I made it `walls: false`, I think it makes sense to be able to sense blood from around a wall. Sensing it in an entirely closed off room, maybe not, but that's not the sort of logic we can currently perform. Could go either way here.
- I figure burrowing creatures can't sense blood, and the blood of a burrowing creature cannot be sensed.
- If the actor has no `health` resource at all (group actors, module-added subtypes) I figure lenience is better. I don't think this will realistically matter much, but figured it was worth mentioning.

### DIFF
--- a/module/canvas/detection-modes/blood-sense.mjs
+++ b/module/canvas/detection-modes/blood-sense.mjs
@@ -1,3 +1,6 @@
+/**
+ * A custom DetectionMode for perceiving blooded creatures that are wounded.
+ */
 export default class DetectionModeBloodSense extends foundry.canvas.perception.DetectionMode {
   constructor() {
     super({
@@ -35,6 +38,7 @@ export default class DetectionModeBloodSense extends foundry.canvas.perception.D
     if ( !health ) return true;
 
     // Otherwise, allow if any health is missing
+    // TODO: return false for non-blooded creatures, once that characteristic is implemented
     return health.value < health.max;
   }
 }
--- a/module/canvas/detection-modes/blood-sense.mjs
+++ b/module/canvas/detection-modes/blood-sense.mjs
@@ -1,0 +1,40 @@
+export default class DetectionModeBloodSense extends foundry.canvas.perception.DetectionMode {
+  constructor() {
+    super({
+      id: "bloodSense",
+      label: "DETECTION_MODES.BloodSense",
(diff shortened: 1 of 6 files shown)


module/canvas/detection-modes/blood-sense.mjs
```diff
@@ -0,0 +1,40 @@
+export default class DetectionModeBloodSense extends foundry.canvas.perception.DetectionMode {
```
aaclayton, line 4: Minimal class docstring please that explains what Blood Sense does and what it provides, mechanically.

module/canvas/detection-modes/blood-sense.mjs
```diff
@@ -0,0 +1,40 @@
...
+    // Cannot smell blood if burrowing, or if target is burrowing
+    if ( source.hasStatusEffect(CONFIG.specialStatusEffects.BURROW) ) return false;
+    if ( target.document.hasStatusEffect(CONFIG.specialStatusEffects.BURROW) ) return false;
+
+    // If actor has no health attribute, allow detection
+    const health = target.actor?.resources?.health;
+    if ( !health ) return true;
+
+    // Otherwise, allow if any health is missing
+    return health.value < health.max;
```
aaclayton, line 42: If we make this a built-in detection mode, I think we may need a built-in "blooded" field on the Taxonomy that articulates whether the creature has blood or not, similar to how the taxonomy tracks whether they are warm-bodied or cool-bodied.

aaclayton (changes requested):
Looks good, couple remarks.

aaclayton (approved):

PR #1449, opened 2026-08-31
[#1441] Fix certain weapon affixes not working properly

### LINKED ISSUE #1441: Some Weapon Affixes do nothing
- All weapon damage Affixes don't add any bonus damage at all
- The Vicious Affix does not add the condition on a critical hit

roth-michael:
Closes #1441 
- Weapon damage affixes' change to `system.damage.bonus` were getting clobbered in `CrucibleWeaponItem#prepareEquippedData`. Now adding to the bonus instead of setting it. I _think_ this should be fine, because as far as I can tell we'll never call this outside derived data prep, so it doesn't need to be idempotent.
- Vicious is looking at `event.weaponItem.id`, which is `null` when cloned without `keepId: true`. A different fix would be to just check `event.weapon._id`, instead, which might be better, but I figure this change will keep us from footgunning in the future as well.

### DIFF
--- a/module/models/action.mjs
+++ b/module/models/action.mjs
@@ -315,7 +315,7 @@ class CrucibleActionEvent {
     if ( this.#weaponItem ) return this.#weaponItem;
     const source = this.#action.actor.items.get(this.weapon._id);
     if ( !source ) return undefined;
-    const clone = source.clone();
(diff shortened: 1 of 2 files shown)


aaclayton (approved):
Thanks!

PR #1451, opened 2026-09-03
[#1313] Fix enrichers being invisible in RollTable draw cards

### LINKED ISSUE #1313: Crucible enrichers not visible in roll table draw cards
Image

XinSysVTT:
Closes #1313

`enriched-content.hazard-check` (and the sibling `.skill-check`/`.passive-check`/`.award`/`.milestone`/`.counterspell`/`.rule` enrichers) color their text with `--color-text-emphatic`, which they never define themselves — it's only supplied by `.themed.theme-dark`, added in `CrucibleChatMessage.onRenderHTML` (`module/documents/chat-message.mjs`) when a message has `flags.crucible` or a `StandardCheck` roll. A native `RollTable` draw message has neither, so an enricher inside a result's description (e.g. `[[/hazard]]`) renders its background/border but not legible text.

Added local fallbacks for `--color-text-emphatic` (and `--color-border` for `.rule`, same gap) directly on the affected selectors in `styles/journal.less`, matching the existing `--color-text`/`--color-frame` values from `variables.less`. Ran `npm run less` to confirm a clean rebuild.

### DIFF
--- a/styles/journal.less
+++ b/styles/journal.less
@@ -163,6 +163,7 @@ enriched-content.counterspell {
   --color-background-hover: #000000;
   --color-border: #000000;
   --color-border-hover: #000000;
+  --color-text-emphatic: #eee8df; // Local fallback; legible outside a themed .crucible card (e.g. RollTable draw)
 
   display: inline-flex;
   height: 1.5em;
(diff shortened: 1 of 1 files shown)


XinSysVTT (commented):
Before: 



After:

aaclayton (approved):
Thanks!

PR #1453, opened 2026-09-04
[#1452] Add `blooded` characteristic to ancestries & taxonomies

### LINKED ISSUE #1452: Implement a "blooded" characteristic on taxonomies & ancestries which, if false, prevents detection via Blood Sense even when wounded

roth-michael:
Closes #1452 
Pretty straightforward for the most part. Things worth noting/double checking:
- Fixed the i18n for temperature hint (it's warm by default for ancestries)
- We only inherit creature category bloodedness either in the case that `blooded` doesn't exist in source (`initial`, this way no existing taxonomies need adjusting) _or_ when modifying taxonomy _via the sheet_. Figure a programmatic update would know to set `blooded` if desired
- The `initial` function is a little bit messy because of the way detail items are handled on an actual actor - sometimes the `source` is actor source data, sometimes it's taxonomy source data
- Added a typedef for `CrucibleCreatureCategory` as it was getting a little crowded
- Should probably do a pass on my determination for bloodedness defaults for each category (undead in particular feels like it varies greatly based on the undead itself. I defaulted to `true` but not sure whether it makes more sense to do that and make changes for `skeleton` for instance, or default to false and make changes for something like a recently raised thrall. I guess probably the latter. But I anticipate potential changes there anyway so leaving as-is for your review

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1059,8 +1059,10 @@
         "label": "Identifier"
       },
       "characteristics": {
+        "blooded.hint": "Whether this Ancestry has some form of blood.",
+        "blooded.label": "Blooded",
         "label": "Ancestry Characteristics",
-        "temperature.hint": "The body temperature of this Ancestry, neutral by default.",
+        "temperature.hint": "The body temperature of this Ancestry, warm by default.",
         "temperature.label": "Body Temperature"
       },
       "movement": {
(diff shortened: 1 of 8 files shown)


aaclayton (approved):
Looks great, thank you!

PR #1457, opened 2026-09-08
Fixes to hooks for proficiencies branch

roth-michael:
Changes:
- Search gets `harmless` to avoid dealing damage
- `_recordConcealmentDC` also marks the effect as always show (technically unnecessary for the case of Stealth, but felt cleaner to do so there than to do a nested loop over events & effects just for Hide)
- Fix Hide hook not properly gathering observer actors
- Fix Search hook not properly gathering hiding actors
- Fix Brutal Display not necessarily working for unimportant Adversaries
- Fix Focal Reach not properly adding to weapon range
- Fix Hide erroneously performing an embedded document deletion during `preActivateAction` rather than recording an `effect` event which deletes the effect in question
- Fix Piercing Bolts clobbering any vulnerability

### DIFF
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -1621,7 +1621,7 @@ export const DEFAULT_ACTIONS = Object.freeze([
     cost: {
       action: 3
     },
-    tags: ["awareness", "subtle"]
+    tags: ["awareness", "subtle", "harmless"]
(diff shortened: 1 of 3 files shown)


aaclayton (approved):
Good fixes!

PR #1458, opened 2026-09-08
Miscellaneous cleanup for proficiencies branch

roth-michael:
Changes:
- Removed a number of unnecessary `_loc`s (things being translated were prelocalized)
- Removed unnecessary jsdoc for Hexed
- Fixed acquisition of `name` and `img` for Hexed
- Used more accurate jsdoc for Staggered
- Expanded each entry in `proficiency.mjs` enums which did not fit entirely on one line such that each property is on its own line to improve readability
- Removed an unnecessary newline in `module/const/spellcraft.mjs`
- Removed unnecessarily verbose comment regarding weapon training talent migrations
- Changed `globalThis._del` to simply `_del`
- Removed a number of unnecessary `{suffix: "0"}` arguments when calling `getEffectId` - by default the return value is right-padded with `0`s to length 16, `suffix` is only necessary if you explicitly want whatever suffix is being provided to be the last `n` characters of the ID (where `n` is the length of `suffix`)
- Removed an unnecessary comment in Orbiting Aegis hook
- Enforced minor grammatical pedantry on `trainingBudget` getter's docstring
- Removed an unnecessary comment in `CrucibleAdversaryActor#_prepareTraining`
- Removed an out-of-place `.` from an inline comment in `CrucibleBaseActor##preparePhysicalDefenses`
- Removed "Skill Preparation" section label in `CrucibleBaseActor` in favor of the new "Training Preparation"
- Amended an incorrect deprecation tag in `CrucibleArchetypeItem.migrateData`

### DIFF
--- a/module/applications/sheets/item-tool-sheet.mjs
+++ b/module/applications/sheets/item-tool-sheet.mjs
@@ -33,7 +33,7 @@ export default class CrucibleToolItemSheet extends CrucibleBaseItemSheet {
   async _prepareContext(options) {
     const context = await super._prepareContext(options);
     context.skillOptions = Object.values(SYSTEM.SKILLS).map(({id: value, label, group}) => {
-      return {value, label, group: _loc(SYSTEM.PROFICIENCY.GROUPS[group].label)};
+      return {value, label, group: SYSTEM.PROFICIENCY.GROUPS[group].label};
     });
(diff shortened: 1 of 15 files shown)


aaclayton (approved):
Thanks for the tidying!

PR #1459, opened 2026-09-08
Other misc cleanup for proficiencies branch

roth-michael:
Felt different enough in scope as to merit its own PR, while the previous misc changes one was almost entirely cleanliness, this one has a little bit of potential mechanical impact.
Changes:
- Rewrote `initializeComponents` to not require a hardcoded array
- Moved the Hero-only `_prepareTraining` logic into `CrucibleHeroActor#_prepareTraining`, leaving `CrucibleBaseActor#_prepareTraining` empty
- Have Talent sheets inherit the same styling as archetype & taxonomy sheets. As far as I can tell there shouldn't be any harmful overlap here, and this is the simplest way of getting proper styling for "Required Proficiencies" without duplicating a chunk of style code

### DIFF
--- a/module/const/spellcraft.mjs
+++ b/module/const/spellcraft.mjs
@@ -629,9 +629,9 @@ export const COMPOSED_SPELL_NAMES = {
  * Rebuilding mutates in place because Talent schema fields captured this object as their choices.
  */
 export function initializeComponents() {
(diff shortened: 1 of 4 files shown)


aaclayton (approved):
Approving and merging, but I think I want to make a slightly more targeted CSS change to scope only the section of rules that `.crucible.talent` needs to share with archetype and taxonomy instead of having it inherit the entire set of rules.

PR #1460, opened 2026-09-08
Genericize `applyMoraleWave`, implement Swashbuckler

roth-michael:
Couple things here worth calling out:
- I decided "nearby allies" can be "within 30 feet" for Swashbuckler
- I decided that "achieve a Critical Hit" means any action-driven critical success targeting _not_ oneself. A Swashbuckler who crits with a spell, or a skill attack, will trigger the morale wave currently
- All 3 currently implemented morale wave talents are "X who can see Y" - we only check sight-blocking _walls_, not detection. Not sure if we'd want to roll that into `wallType` being "sight" or do something else, for now I've not addressed it
- Because we're now iterating over `eventsByActor` rather than `eventsByTarget` (necessary in order to accommodate the acting token being a source of the wave), disallowing downstream waves to occur as a result of side effects is something that must be handled in the `test` parameter. What I mean by that is that previously, Brutal Display could proc, damaging the morale of nearby enemies. If any of them became Broken, that would have no impact on Rout firing, because none of them were targets of the original action. Now, that may not be the case. I've added a check to both Brutal Display's and Rout's `test`s to continue excluding non-targets from consideration as a wave source, but this is something worth keeping in mind for future implementations
- Line 61 (now 71) of `talent.mjs` was a mostly-silent bug, but was accessing `document` on what was already a Token Document

### DIFF
--- a/_source/talent/Swashbuckler_swashbuckler0000.yml
+++ b/_source/talent/Swashbuckler_swashbuckler0000.yml
@@ -5,12 +5,12 @@ system:
   description: >-
     <p>Your bravado inspires your allies and cowes your enemies. When you Dodge
     or Parry a Strike, the attacker loses Morale equal to your Presence
-    attribute. When you achieve a Critical Hit, all nearby allies who can see
-    you gain Morale equal to your Presence attribute.</p>
+    attribute. When you achieve a Critical Hit, all allies within 30 feet who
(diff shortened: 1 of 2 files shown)


aaclayton (approved):
Good work, thanks for this cleanup, and I'm glad that it gets us Swashbuckler unlocked!

PR #1463, opened 2026-09-08
add ability tooltip to actor sheet, hero creation, talenttree

### LINKED ISSUE #1462: add ability tooltip to actor sheet, hero creation, talenttree
it is a little unclear what abilities or attributes actually affect (derived attributes) and on the talenttree the colors are only indirectly related to retrieving the attribute

improvement:

* tooltip taken from journal description upmost part without the example and the 0 value explanation so it does not become too big
* for talenttree a little more intrusive since it requires a new tooltip kind/template. potentially could be converted to node to refactor but i think tooltip is worth

Target: general better accessibility and rules available without looking into game description via tooltips

PR inc in a bit

Plushtoast:
Fixes #1462

Adds tooltip for abilities to:

* actor sheet
* hero creation
* talenttree

tooltip is taken from game journal

grafik

grafik

grafik

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -5,6 +5,14 @@
       "Resilience": "Widerstand",
       "Speed": "Geschwindigkeit"
     },
+    "TOOLTIPS": {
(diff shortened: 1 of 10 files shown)


aaclayton (approved):
I want to refactor this to use the new `RULES` framework, but rather than push that back to you I can handle it as a follow-up commit.

PR #1466, opened 2026-09-08
add tooltip to group actor for rest and recover

### LINKED ISSUE #1465: add tooltip to group actor for rest and recover
it is unclear via the group actor sheet what is the difference for rest and recover

a tooltip can clarify implications rules and consequences of those buttons

Plushtoast:
Fixes #1465

grafik
grafik


aaclayton (approved):
Will make a follow-up commit to integrate with `RULES`.

PR #1468, opened 2026-09-09
enhance requestRoll actors to toggle between online party members (default) and all party members in roll request

### LINKED ISSUE #1467: enhance requestRoll actors to toggle between online party members (default) and all party members in roll request
I see myself using the add group members to a roll request a lot.

what is annoying is that on a group of maybe 5 players one player is always offline but still appears in the roll request. this leads to the following inconveniences:

* need to remove offline actors manually
* need to add each actor separatedly to avoid offline actors
* roll with offline actors and need to modify the chatcard with skip
* another player rolls for offline actors if they have permissions

I think a better default would be if any roll defaults to online members only and the add group actor button acting as a toggle between online members and all members

in my experience this is the better default because offline actors just either wait in the camp or stand in a corner etc and are not likely to walk into hazards or solve riddles as the gamemaster

Plushtoast:
Fixes #1467

Also refactors the party retrieval into a static method getParty which inclludes the "party has no members" warning and thus reducing code duplication

Adds everyone to a roll if nobody is active mainly for debugging or testing issues for gm and dev

### DIFF
--- a/module/dice/hazard-dialog.mjs
+++ b/module/dice/hazard-dialog.mjs
@@ -272,18 +272,15 @@ export default class HazardDialog extends ActionUseDialog {
   /* -------------------------------------------- */
 
   /**
-   * Add the configured party members to the targets list.
+   * Toggle party members in the targets list, online-first.
    * @this {HazardDialog}
    * @param {Event} _event
    */
   static async #onTargetsParty(_event) {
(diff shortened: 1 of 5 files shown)


aaclayton (approved):
Good improvement, thanks @Plushtoast

PR #1471, opened 2026-09-10
Add preliminary linkage of effect to originating action

roth-michael:
A prerequisite to having auras reference their originating action for purposes like targeting. As mentioned in TODOs, there's definitely room for improvement here, but for the time being this is sufficient for our purposes and we can stick with it until we've got a better permanent plan worked out.


module/documents/actor.mjs
```diff
@@ -1217,20 +1217,24 @@ export default class CrucibleActor extends Actor {
...
-  async _applyActionEffects(effects, {reverse=false, commit=true}={}) {
+  async _applyActionEffects(effects, {reverse=false, commit=true, originAction}={}) {
     if ( !effects.length ) return;
     const toCreate = [];
     const toUpdate = [];
     const toDelete = [];
     for ( const effectData of effects ) {
+
+      // TODO: Store more action information than just identifier, likely on data model rather than flags
+      if ( originAction ) foundry.utils.setProperty(effectData, "flags.crucible.originAction", originAction.id);
```
aaclayton: I think this should be moved to `CrucibleAction##recordEffectEvents` which already writes the `origin` for the same action.

aaclayton: A second problem here that we might possibly be updating a pre-existing effect, which would result in overwriting a predefined `originAction` (hypothetically). Make sure to assign this only on new effect creation and then any further updates that modify the origin would be viewed as intentional.

module/models/effect-base.mjs
```diff
@@ -60,4 +60,23 @@ export default class CrucibleBaseActiveEffect extends foundry.data.ActiveEffectT
...
   }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Retrieve the action which caused the creation of this effect.
+   * TODO: Once storing more complete action data, make this more robust; this is mainly a placeholder proof of concept
+   * @param {object} [options]
+   * @param {CrucibleActionContext} [options.actionContext] Specific action context, if any
+   * @returns {CrucibleAction|undefined}
```
aaclayton, line 76: I think you need type-only imports for `CrucibleAction` and `CrucibleActionContext` here.

module/models/effect-base.mjs
```diff
@@ -60,4 +60,23 @@ export default class CrucibleBaseActiveEffect extends foundry.data.ActiveEffectT
...
+   */
+  getOriginAction({actionContext={}}={}) {
+    const actor = fromUuidSync(this.parent.origin);
+    if ( !(actor?.system instanceof crucible.api.models.CrucibleBaseActor) ) return;
+    const actionId = this.parent.flags?.crucible?.originAction;
+    if ( !actionId ) return;
+    return actionId.startsWith("spell.")
```
aaclayton, line 84: I think reconstructing an origin action should use the same `{lazy: true}` that we use when cloning an action - sharing the existing `usage` context instead of recreating it. 

See the precedent in `module/models/action.mjs:3576`

aaclayton (changes requested):

PR #1472, opened 2026-09-10
Implement logic for "Crucible Action" region behavior

roth-michael:
Next piece of logic required for auras. Do not expect the config to look pretty, or even be usable at this point in truth, since it's just core auto-generating the field groups. At most you can define the name and watch it obey the "once per turn" logic.

I also did a small cleanup in the hazard behavior, making its valid events a private static.


module/models/behavior-action.mjs
```diff
@@ -0,0 +1,85 @@
...
+    const actionContext = {actor: sourceActor, region: this.parent.parent};
+    const originEffect = await fromUuid(this.origin);
+    if ( originEffect ) {
+      const originAction = originEffect?.system.getOriginAction({actionContext});
+      const validTargets = new Set(originAction?.acquireTargets().keys() ?? []);
+      if ( !validTargets.has(actor) ) return;
+    }
+
+    // If once per round and already done this round, skip
+    if ( this.oncePerRound && game.combat && (this.affectedActors[actor.uuid] === game.combat.round) ) return;
```
aaclayton, line 71: Caution, if `game.combat` is `null` we would end up performing the action for every `tokenMoveWithin` action. We need to think carefully about what should happen outside of a tracked combat encounter. Do you have any good ideas on that?

Also I'm not sure whether it's a risk that `actor` might not be a participant in `game.combat` - I suppose that doesn't strictly matter, we track `affectedActors` either way.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,85 @@
...
+    // Otherwise, perform action
+    const action = new crucible.api.models.CrucibleAction(this.action, {
+      actor: sourceActor,
+      usage: {forcedTargets: [actor]}
+    });
+    action.use({dialog: false});
+
+    // If once per round, track that the targeted actor has been affected this round
+    if ( this.oncePerRound && game.combat ) {
+      await this.parent.update({"system.affectedActors": {[actor.uuid]: game.combat.round}});
```
aaclayton, line 82: If we are already making a database round-trip, should we use the opportunity to force replace `affectedActors` down to the minimum set of entries we still need to track for this round of combat?

It seems to me there is some risk that we could accumulate old entries here, which could be some arbitrary round number for an actor - possibly even an old combat encounter that is no longer the current one. 

There might even be a risk that some actor is tracked as having already been affected during "round 3", but it was round 3 of some prior combat encounter, and so they get incorrectly skipped during round 3 of the current encounter? Cleaning the record might be a solution for avoiding that.

Alternatively we could make the data structure more verbose and store a `{combatId, round}` tuple for each actor. 

I think the *maximally correct* treatment would be:

1. We need to store an array of `{combatId, round}[]` per actor - an actor can belong to multiple tracked combat encounters.
2. We need to prune the record to eliminate no-longer-relevant entries either in response to some event (like `combatStart`) OR as a side effect of making any other update.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,85 @@
...
+
+    // If once per round and already done this round, skip
+    if ( this.oncePerRound && game.combat && (this.affectedActors[actor.uuid] === game.combat.round) ) return;
+
+    // Otherwise, perform action
+    const action = new crucible.api.models.CrucibleAction(this.action, {
+      actor: sourceActor,
+      usage: {forcedTargets: [actor]}
+    });
+    action.use({dialog: false});
```
aaclayton, line 78:
```suggestion
    await action.use({dialog: false});
```

aaclayton (changes requested):
Good next step - a couple of key structural issues to work on next. Thanks @roth-michael!

module/models/behavior-action.mjs
roth-michael: I suppose if we do a once-per-round outside of combat it might make sense to record a round of `-1` or similar, then check against that. Would effectively mean any "one per round" could only happen once outside of combat too. Given that, in an aura context, these tend only to affect resources, _and_ out of combat resource recovery is trivial if you've got 10 minutes, I think this might be appropriate. Let me know what you think.

module/models/behavior-action.mjs
roth-michael: I’m definitely down for pruning of any lte-current-round keys. Tracking different combats I'm les sure about. If somehow two actors are both in two separate combats, doing a per-actor-per-combat check would mean an actor could get hit twice in the same 10 second (1 round) span by the same effect.

My inclination is that multi-combat consideration for this type of behavior is something we can largely ignore. Particularly if we prune. The one collision that we might hit is "enemy was hit on round 3, combat immediately ends, new combat immediately begins." 

I’m not sure how we'd address that. For the case of an effect-linked behavior, we could iterate over combatant effects on combat start and/or end and clear their dependent regions' `affectedActors`. Don't think we can do something similarly clever for non-effect-linked behaviors. For a scene-linked combat we could, but that's not a great solution because not all combats are scene linked.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,85 @@
...
+        expandKeys: false,
+        validateKey: uuid => {
+          const {id, type} = foundry.utils.parseUuid(uuid);
+          return (type === "Actor") && foundry.data.validators.isValidId(id);
+        }
+      }),
+      events: this._createEventsField({events: this.#VALID_EVENTS, initial: ["tokenEnter", "tokenTurnStart"]}),
+
+      // Whether this should apply to a given actor only once per round, on every trigger
+      oncePerRound: new fields.BooleanField({initial: true, required: true, nullable: false}),
```
aaclayton: I wonder if instead of `oncePerRound (boolean)` we should design around `frequency (string)` with possible values like `"ever"|"combat"|"round"|"turn"`. Something like that might inform the frequency with which we need to expire the record-keeping?

module/models/behavior-action.mjs
aaclayton: I think "once outside of combat" feels like an okay policy. Would that reset when a combat encounter occurs? So for example you could be affected by a region before combat, then again during combat (once per round), and then again after combat ends?

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,115 @@
...
+        round: game.combat?.round ?? -1,
+        turn: game.combat?.turn ?? -1
+      }}});
+
+      // If in combat, record this behavior to the current Combat so that affectedActors is cleared on combat deletion
+      if ( game.combat ) {
+        const combatBehaviors = game.combat.getFlag("crucible", "trackedActionBehaviors") ?? [];
+        if ( combatBehaviors.includes(this.parent.uuid) ) return;
+        combatBehaviors.push(this.parent.uuid);
+        await game.combat.setFlag("crucible", "trackedActionBehaviors", this.parent.uuid);
```
aaclayton: I think this line is supposed to write `combatBehaviors`, not `this.parent.uuid`.
```suggestion
        await game.combat.setFlag("crucible", "trackedActionBehaviors", combatBehaviors);
```

module/documents/combat.mjs
```diff
@@ -103,7 +102,29 @@ export default class CrucibleCombat extends foundry.documents.Combat {
...
-    if ( actorUpdates.length ) Actor.updateDocuments(actorUpdates);
+    if ( !isGM ) return;
+    const batchOperations = [];
+    if ( actorUpdates.length ) batchOperations.push({
+      action: "update",
+      documentName: "Actor",
+      updates: actorUpdates
+    });
+
+    // Flush the affectedActors tracking of any Crucible Action behaviors which fired during this combat
```
aaclayton, line 113: I would like the logic for how to clean `affectedActors` to live on the behavior that owns that piece of data, so let's move the inside of this loop to be an instance method of the behavior.

I think *for now* we should also reset the tracker for "which actors were affected while out of combat" (the -1 values) IFF the actor was present in this combat that is ending. Thoughts?

aaclayton (changes requested):

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,115 @@
...
+    CONST.REGION_EVENTS.TOKEN_ROUND_END
+  ];
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    const {id, name, img, description, effects, tags} = crucible.api.models.CrucibleAction.defineSchema();
+    const frequencyChoices = {
```
aaclayton, line 28: I think we have a real use case for a `"once"` frequency that would cause the action to only fire once ever - the "Steel-Jaw Trap" item would be a good candidate for that.

module/models/behavior-action.mjs
roth-michael: Whoops. And my naive testing was working because `["something"].includes("something")` and `"something".includes("something")` both return true. Good catch.

module/documents/combat.mjs
roth-michael: Makes sense for both. Can always revisit the tracker reset as we end up with more actual use cases.

module/models/behavior-action.mjs
```diff
@@ -0,0 +1,155 @@
...
```
aaclayton, line 28: I think we should hoist these to be a private static, as `#VALID_EVENTS`.

roth-michael: I actually do make them static in a later PR, but not private so that they can be used as `choices` in the Action schema.

PR #1474, opened 2026-09-10
Implement custom config app for "Crucible Action" region behavior

roth-michael:
With this, the Crucible Action behavior is fully functional standalone. You can create a behavior, configure an action to be executed by that behavior (complete with most of the goodies you'd expect from an action - tags, effects, an identifier if you want it to point at some hooks you've got), specify which actor should be rolling as the "attacker" and which events should trigger it, and it'll roll that action against tokens that trigger the events as specified.


styles/applications.less
```diff
@@ -144,6 +144,10 @@
           --input-background-color: var(--color-cool-5-50);
           --input-text-color: var(--color-text-primary);
         }
+        &.increased-contrast {
+          background: var(--color-frame-bg);
+          border: 1px solid var(--input-border-color);
+        }
```
aaclayton: Seems unrelated

styles/applications.less
```diff
@@ -316,6 +320,18 @@
       }
     }
   }
+
+  /* -------------------------------------------- */
+  /*  Region Behaviors                            */
+  /* -------------------------------------------- */
+
+  &.region-behavior-config .sheet-body.scrollable {
```
aaclayton: When we are using the `.scrollable` class, the right margin and padding are controlled by a variable, aren't they? Other properties like `scrollbar-gutter: stable` should also come for free?

module/models/action.mjs
```diff
@@ -612,17 +612,37 @@ class CrucibleActionTags extends Set {
...
-    });
+      // Limit allowed effect scopes
+      const effectScopes = SYSTEM.ACTION.TARGET_SCOPES.choices;
+      delete effectScopes[SYSTEM.ACTION.TARGET_SCOPES.NONE]; // NONE not allowed
 
-    // Limit allowed effect scopes
-    const effectScopes = SYSTEM.ACTION.TARGET_SCOPES.choices;
-    delete effectScopes[SYSTEM.ACTION.TARGET_SCOPES.NONE]; // NONE not allowed
+      // Must manually set label, as these'll exist both here and under Action Region Behavior schema
+      const makeLabel = fieldName => _loc(`ACTION.FIELDS.effects.element.${fieldName}.label`);
```
aaclayton: This doesn't seem like the right approach to me, labels and hints should get applied via data model prelocalization and there shouldn't be any labels directly supplied at the `defineSchema()` step at all.

module/applications/_module.mjs
```diff
@@ -2,6 +2,7 @@
 export * as elements from "./elements/_module.mjs";
 
 // Config Apps
+export {default as CrucibleActionBehaviorConfig} from "./config/action-behavior-config.mjs";
```
aaclayton: It's very verbose, but let's commit to the bit of just appending "Config" to the data model name, so this would be `CrucibleActionRegionBehaviorConfig`

templates/sheets/region-behavior/action-behavior-config.hbs
```diff
@@ -0,0 +1,47 @@
+<section class="standard-form scrollable sheet-body">
+    {{#each fields as |outer|}}
```
aaclayton: You're being clever by defining the fields here through iteration, but it comes with some cost (more logic in the template, mainly). I think I'd prefer just explicit declaration, even if its a few lines longer.

templates/sheets/region-behavior/action-behavior-config.hbs
```diff
@@ -0,0 +1,47 @@
...
+    {{else}}
+    {{formField outer.field value=outer.value localize=true model=../model}}
+    {{/if}}
+    {{/each}}
+    <fieldset>
+        <legend>{{localize "TYPES.RegionBehavior.crucible.action"}}</legend>
+        {{formGroup systemFields.origin value=source.system.origin disabled=true}}
+        {{formGroup systemFields.actor value=source.system.actor}}
+        {{formGroup systemFields.frequency value=source.system.frequency localize=true}}
+        <fieldset>
```
aaclayton: I wonder if there is any good way that we can get these inner action fields in some way that is reusable. Not critical, just musing. We are having to reinvent some specific styling decisions here that it would be nice if we could inherit from elsewhere.

aaclayton: I'm wondering if we could have a UX that gives you a simple choice between:

A. Specify the identifier of a pre-existing Action that belongs to the `actor` for the behavior.

B. Configure a custom action that is embedded within the behavior and exists outside of the Actor who performs it.

If you choose route B, I wonder if we should just open a simplified variant of the `CrucibleActionConfig` instead of trying to present these fields here inline.

aaclayton (changes requested):

aaclayton (approved):

PR #1475, opened 2026-09-10
Embed pre-configured region behavior data on actions, tie to creation of persisted regions

### LINKED ISSUE #610: V14 Aura Regions

### LINKED ISSUE #611: V14 Aura Behaviors

roth-michael:
Last piece of the puzzle (not including updating some existing actions to use this new functionality).

Copying over a bit of text from #1008 
> - I split out the "use some self event to record the existence of a persistent region" logic from the "update region visibility & behavior disabled status (& now, setting the origin uuid) so that the first bit happens prior to effect creation, and the latter bit post effect creation. Reason being: Events will trigger potentially immediately upon enabling the region behavior. Therefore the origin effect must exist at that point to be referenced. As a consequence of this, the previous logic (which updated region visibility depending on reversal) has been amended such thot, during reversal, nothing happens to the region. This is because a persistent region will, by this point, be slated for deletion already as a downstream effect of deleting the tracking active effect, so odds are high (if not certain?) that trying to call update on this doomed region will throw an error.
>   - Part of this change involved moving the retained check up prior to the "just set the first self effect to track it" logic, which means a hook could theoretically specify a specific self-effect (in the event that there are multiple) to be the region-tracking one. Also means down the line we could potentially have an action that creates an aura on somebody else and has the effect which tracks it on them too. Dunno if we ever will, but we could with this logic, I think.

And then, as this closes #610 and closes #611, some comments on what it does _not_ address from those issues:
- "Automatically disable the aura if the associated `ActiveEffect` is disabled or suppressed" see https://github.com/foundryvtt/crucible/issues/610#issuecomment-4746016547
- "Implement a region behavior type for Aura actions that directs region events back to the originating Action for handling via action hooks." I think the scope of the behavior has changed, so can still call this complete, but rather than directing to the original Action's hooks, it instead has an "embedded" Action which may have its own hooks


aaclayton (approved):
Merging with some follow-up fixes in mind.

PR #1476, opened 2026-09-10
Implement Lifebloom and Font of Life as proper auras

roth-michael:
Two standout "should've been auras" actions here converted to in fact be auras.


roth-michael:
Closing this in favor of a slightly broader PR addressing other "Crucible Action" behavior-benefitting actions.

PR #1481, opened 2026-09-11
Implement the Decouple Shadow action to summon a caster-sized, leashed shadow for the Rune of Illusion

### LINKED ISSUE #1426: The Rune of Illusion action for decoupling a shadow should provide a stated maximum range.

XinSysVTT:
Closes #1426

### Problem

The Seeming action of the Rune of Illusion allowed a character to "decouple their shadow from their
form, specifically controlling its movements separately from your own", but the effect was purely
narrative: activating the action spent the Action and Focus cost and posted a chat card, after which
the shadow existed only in the fiction. There was no board presence to move, flank with, or interact
with, and no range limitation beyond the prose.

### What this PR adds

- A new **Decouple Shadow** action on the Rune: Illusion talent (3 Action, 1 Focus, 1 free hand).
  Placement uses the existing summon-target pipeline: the shadow manifests at a point within 30 feet,
  clamped by `clampSummonPlacement` exactly like the Warden's Ancestral Grove.
- A new **Shadow** creature in the `summons` compendium (illusory construct, no attacks, base size 3,
  stride 8) with a generic token asset (should be replaced with something better) (`assets/tokens/creations/Shadow.webp`) and a small
  darkness light source (negative light, dim 4). Sight is disabled; the shadow does not see. I did not make it full darkness unsure if this was the plan? 
- Two new optional capabilities for the summon action tag, both opt-in and inert for existing summons:
  - `leash` — a maximum distance in grid units the summoned creature may move from its summoner.
    The created token is stamped with `flags.crucible.leash` (`sourceTokenId`, `itemUuid`,
    `distance`), and a handler bound to the `preMoveToken` hook
    (`CrucibleCanvasMovement#enforceSummonLeash`) blocks any move that would end beyond that
    distance, warning the player.
  - `matchCasterSize` — the summoned creature's token footprint is sized to its summoner's. The
    summon tag's `initialize` copies the caster's token dimensions into the summon `tokenData` so
    the placement preview is caster-sized.
- Re-using the action displaces the caster's previous leashed summon (one leashed summon per
  summoner per action), matching the "maintain this illusion" wording. Displacement is scoped per
  summoner token + item and only deletes tokens the user may delete.
- Seeming's shadow bullet now points to the Decouple Shadow action instead of describing a purely
  narrative effect.

Will upload images in a second


XinSysVTT:
{D6A2572E-CEA5-4C58-BF67-C95580C2366E}


{7BB8813C-90D2-4F24-A3B1-44591495B8A6}

{4A94A042-BECC-48FD-8A2D-515522E39C80}

aaclayton:
Sorry, I think this PR is doing a few too many things at once. If we were going to consider it, we would need to break it up into some component parts which can be individually merged as justified low-level components that build-up to the eventual functionality we want for the shadow.

Are you interested in taking that on? Or should I close this PR?

PR #1485, opened 2026-09-12
[#1256] Use an Emanation region for Pulse targeting

### LINKED ISSUE #1256: Switch actions that use the "Pulse" target type to produce an Emanation region, rather than a Circle
We had some discussion of this in the dev channel, spawning from **Dervish**'s Whirlwind action, which should be "you attack every enemy within range." Currently it's modeled as a pulse, and pulses use circular regions, which means inaccuracies at corners. If pulses were instead emanations, this would work fine.

That's not the only possible solution for Whirlwind, though, so it's worth considering whether pulses _should_ be emanations. I think I'm leaning towards "yes" - currently we already do the "add half the width of the target to the radius" logic ourselves, which would be main benefit 1 of using an emanation, but I also think using an emanation is logically consistent with what I'm imagining a "pulse" to actually represent, and think if the pulse targeting type had been first created in a post-v14 world, it would've used emanation.

If you agree with all the above, the only thing remaining would be to consider whether this would impact any existing actions in a way we don't want.

XinSysVTT:
Pulse-targeted actions currently use a **Circle** region anchored on the acting token's
*center*, with half the token's width added to the radius to compensate for the caster's
footprint. As discussed in #1256, this produces two inaccuracies:

- The circle bulges past the token's edges on the orthogonal axes while **clipping through
  the corners of the token's own footprint** — large creatures don't even cover themselves
  at the corners.
- Coverage near the diagonal corners doesn't match "every enemy within range" (e.g.
  Dervish's Whirlwind), because the radius is measured from a center point instead of from
  the creature's actual space.

This PR switches the `pulse` target type to use an **Emanation** region, the same shape the
`aura` target type already uses. The radius is measured from the token's footprint, so the
token's size is accounted for natively (no more manual `addSize` half-width offset) and the
area follows the creature's actual shape. If the pulse targeting type had been created in a
post-v14 world, this is the shape it would have used.

### Changes

- `module/const/action.mjs` — `TARGET_TYPES.pulse.region`: `circle` → `emanation`,
  `addSize: true` → `addSize: false` (mirroring the `aura` config).
- `module/dice/action-use-dialog.mjs` — preserve the pulse's vertical reach by extending the
  region elevation range by the pulse radius in the target-type switch (previously done by
  the `circle` shape branch).
- `lang/en.json` — generalize the `AuraNoToken` warning to "area-target" actions, since
  Pulses now share the Emanation placement path.

### Before / After

Scenario: a size-4 (4×4 pace) automaton uses a pulse with range 2 against dummy tokens.
Dummies sit orthogonally adjacent/2 out, diagonally adjacent/2 out, and at the two
discriminator positions just beyond the footprint corners.

*Before — Circle region: the area is centered on the token center, overreaching the edges
orthogonally while clipping the token's own corners. Only 3 targets acquired, and the
corner dummies (within 2 paces of the creature's corner) are missed.*

*After — Emanation region: the area hugs the token footprint with a uniform 2-pace reach
around every edge and rounded corners. The corner dummies are now correctly acquired
(5 targets).*

Side-by-side comparison: 
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -431,7 +431,7 @@
         }
       },
       "AlreadyConfirmed": "The action against you has already been confirmed and can no longer be reacted to.",
-      "AuraNoToken": "You cannot use an Aura-target action without having a token on the scene!",
+      "AuraNoToken": "You cannot use an area-target action without having a token on the scene!",
       "BadStatus": "You may not perform {action} while {status}.",
       "CannotAffordCost": "{name} has insufficient {resource} to use {action}!",
       "CannotAffordHands": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -431,7 +431,7 @@
         }
       },
       "AlreadyConfirmed": "The action against you has already been confirmed and can no longer be reacted to.",
-      "AuraNoToken": "You cannot use an area-target action without having a token on the scene!",
+      "AuraNoToken": "You cannot use a self-emanating action without having a token on the scene!",
       "BadStatus": "You may not perform {action} while {status}.",
       "CannotAffordCost": "{name} has insufficient {resource} to use {action}!",
       "CannotAffordHands": {
--- a/module/const/action.mjs
+++ b/module/const/action.mjs
@@ -110,9 +110,9 @@ export const TARGET_TYPES = defineEnum({
   pulse: {
     label: "ACTION.TARGET_TYPES.Pulse",
     region: {
-      shape: "circle",
+      shape: "emanation",
       anchor: "self",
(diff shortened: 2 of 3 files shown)


roth-michael:
Looks like an easy win. Only thing I'd note is the new localization string isn't strictly accurate, as any region is an "area-target" action. Probably specifying a self-emanating action would make more sense.

aaclayton (approved):
Willing to give this a try and see if i hate it. I'm somewhat attached to the circle, but this is more mechanically correct.

Issue #993, opened 2026-05-07
Add support for Cursed Affixes


anathemamask:
These affixes would be the inverse of the current affixes, causing negatives on the character. 

Also from the conversation that led to this issue being created:
> Atropos: Maybe even cases where a cursed affix looks like a good affix until you invest the item and realize that it's fucked up. No particular system for that yet though 

Some use cases: 
- Invert the <Damage Type> Resistance benefit, so that it takes away resistance for the affected character.
- Invert the <Creature type> Bane affix so that it gives banes instead of boons.
- Invert the Potency affix so that it deals less damage than the weapon normally would

+

XinSysVTT:
adding a comment as I tried a idea

PR #1486, opened 2026-09-12
993 cursed affixes

XinSysVTT:
Cursed affixes are the inverse of their regular counterparts: they hinder the bearer instead
of helping them. This implements the mechanism discussed in #993, including the "looks like
a good affix until you invest the item" fantasy — a cursed affix is embedded, named, and
capacity-charged exactly like its regular counterpart, but its benefits are inverted once
the item is invested.

### Mechanism

A cursed affix is any affix effect with the new `system.cursed` flag set. Cursed variants
**share the identifier** of the regular affix, so each module hook in
`crucible.api.hooks.affix` reads the flag from the applied effect at runtime and inverts
itself — every invertible affix family gains cursed support automatically, with no duplicate
hook code:

- **Numeric bonuses** (damage, resistance, defenses, health/morale, reach, weapon potency)
  apply negatively via the new `AffixData#sign` helper (`1`, or `-1` when cursed).
  *Examples: a cursed Acid Resistance lowers Acid resistance by 3 per tier; a cursed
  Resistance can push a character into vulnerability.*
- **Best-of bonuses** (rune potency, skill enchantment) apply a negative best-of instead of
  a positive one.
- **Bane affixes** inflict **Banes** on Strikes against the targeted creature category
  instead of granting Boons.
- **Threshold affixes** invert direction: Keen makes criticals *harder*, Reliable makes
  critical failures *easier*, Nonchalant/Unshakeable make critical hits against the wearer
  *easier*, Mending/Rallying raise the healing thresholds.
- **Luminary** inflicts Banes on composed spells instead of granting Boons.
- Affixes with no meaningful inverse (Returning, damage-type Conversion, Spellcraft
  knowledge grants, Vicious, Luminous) ignore the flag.

`StandardCheck` now applies **negative enchantment** values to the dice formula (previously
values ≤ 0 were silently dropped), so cursed potencies genuinely reduce rolls.

### UI & data

- The affix configuration sheet gains a **Curse** section with the `cursed` checkbox and the
  header shows a **Cursed** tag.
- Item sheets reveal cursed affixes to **GMs only** (a red skull indicator), keeping the
  surprise intact for players until the item is invested and the curse takes hold.
- Three example cursed affixes ship in the Affixes compendium, one per use case from the
  issue: **Cursed Acid Resistance** (inverted Resistance), **Cursed Beast Bane** (inverted
  Bane), and **Cursed Potency** (inverted Potency). Their adjectives match their regular
  counterparts (e.g. "Acid-Warding", "Power"), so cursed items name deceptively normally.

### Testing

- Live-verified on Foundry VTT v14 build 367 against an invested item on an unlinked token:
  - Cursed Acid Resistance (tier 2): Acid resistance fell by exactly 6 (resistance bonus 0 → −6).
  - Cursed Beast Bane (tier 2): Strikes vs. a Beast-taxonomy actor gained `banes.beastBane`
    (2 banes) instead of boons.
  - Cursed Potency (tier 2): weapon attack enchantment went negative (derived +1 − 2 = −1)
    and the attack formula rendered the subtraction (`… + 2 + 3 − 2`).
  - Procedural naming composed normally from the cursed adjective ("Beast-Bane Spear of
    Power"), with no visible curse hint in the item name.
  - Affix capacity validation, item name composition, and non-cursed affixes are unaffected.
- Regular (non-cursed) affix behavior is unchanged: the flag defaults to `false` and every
  hook's non-cursed branch is identical to its previous implementation.

- Cursed affixes currently consume prefix/suffix capacity at full tier value like regular
  affixes. If cursed affixes should instead be "free" (hidden cost of the curse), that's a
  one-line change in the capacity accounting — feedback welcome.
- A natural follow-up would be enchanting-workflow integration (e.g. a chance for
  enchanting rolls to produce a cursed variant), or a way to identify curse items, remove the cursed item from equipment etc.

### DIFF
--- a/_source/affixes/Cursed_Acid_Resistance_acidResCursed00.yml
+++ b/_source/affixes/Cursed_Acid_Resistance_acidResCursed00.yml
@@ -1,0 +1,49 @@
+name: Cursed Acid Resistance
+type: affix
+duration:
+  value: null
+  units: seconds
+  expiry: null
+  expired: false
+system:
+  adjective: Acid-Warding
+  identifier: acidResistance
+  changes: []
+  cursed: true
+  affixType: prefix
+  tier:
+    min: 1
+    max: 3
+    value: 1
+  itemTypes:
+    - armor
+    - accessory
+_id: acidResCursed00
+img: icons/magic/death/undead-skeleton-energy-green.webp
+disabled: false
+start: null
+description: >-
+  <p>A mockery of true enchantment. To all inspection this affix appears as a
+  genuine <strong>Acid-Warding</strong> blessing, but once the item is
+  <strong>Invested</strong> the curse takes hold: <strong>Acid</strong>
+  resistance is <em>reduced</em> by 3 for each <strong>Tier</strong> of this
+  prefix, leaving the bearer vulnerable to acid.</p>
+origin: null
+tint: '#ffffff'
+transfer: false
+statuses: []
+showIcon: 1
+folder: ye7FSz6vOcgm2RY8
+sort: 500000
+flags: {}
+_stats:
+  coreVersion: '14.359'
+  systemId: crucible
+  systemVersion: 0.9.1
+  createdTime: 1775502000000
+  modifiedTime: 1775502000000
+  lastModifiedBy: WRGzDgYAt3ZuX6TU
+  compendiumSource: null
+  duplicateSource: null
+  exportSource: null
+_key: '!effects!acidResCursed00'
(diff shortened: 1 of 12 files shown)


aaclayton:
Good prototype, and gave me some inspiration for how to handle this, but I've taken it in a bit of a different direction in commits `52be51cb65ec2281ddd7e3bb22171ec54ab0392b` through `63f6389e3096a573f95aea31ac736742fbac874f`

Issue #452, opened 2025-12-05
Create a `Maintain` action used to maintain `ActiveEffect` documents which should otherwise expire


aaclayton:
Carry-on from https://github.com/foundryvtt/crucible/pull/443

I think a dialog here is okay *for now* but I'd like to see a different solution in the longer term.

1. I think perhaps you should only be allowed to have one (or none) maintained effect at a time, and I think `Maintain` might be a general action you have available to your character while you have a maintained effect. I think ideally we would prompt you to use the `Maintain` action at the start of your turn.

2. This would give us a place to hook in talents that have to do with maintenance by allowing them to modify the cost or other properties of the Maintain action, so we could write mechanics like "When you maintain an effect ... {X} other thing happens".

3. It might be useful during actor prep to identify something like the singleton `maintainedEffect` (or none) rather than iterating over all effects in search of one with a maintenance cost.

4. Using another action which has a maintained effect could prompt you with a dialog confirmation warning, forcing you to acknowledge that your previously maintained effect will be lost.

All just brainstorming, I think a Dialog here is *okay for now*, just not where I want to end up long-term.

PR #1487, opened 2026-09-12
[#452] Add a Maintain action for sustaining maintained effects

XinSysVTT:
### How it works

- **Natural expiry.** Effects created by actions with the `maintained` tag now carry a real
  duration (2 rounds, expiring at the start of the Actor's turn) alongside their maintenance
  cost. A maintained effect persists through the round it was cast and the following round,
  giving the player one full turn to sustain it before it expires naturally.
- **Singleton maintained effect.** Actor preparation resolves `actor.maintainedEffect` (or
  null) once during prep — the issue's "singleton maintainedEffect" — so nothing needs to
  iterate over all effects hunting for maintenance costs.
- **The Maintain action.** A general action that only appears while a maintained effect
  exists, with its focus cost drawn from that effect's maintenance cost. Confirming it
  re-anchors the effect's duration to the current round, deferring expiry. Since it is a
  real action, talents can now hook `prepareAction` / `useAction` / `confirmAction` against
  the `maintain` action id to modify its cost or add riders ("when you maintain an effect…").
- **Singleton enforcement + replacement warning.** Using another action with the
  `maintained` tag while an effect is already maintained prompts a confirmation ("Using X
  will end your maintained Y"); on activation the old effect ends.
- **Turn-start cleanup.** The per-effect DialogV2 confirm and the automatic focus
  affordability deletion are removed from `#prepareTurnStartConfig`; an unmaintained effect
  simply expires through the standard duration pipeline.

### Testing

Live-verified on Foundry VTT v14 build 367 with a combat running:

- With no maintained effect, the Maintain action is absent from the actor's actions.
- Creating a maintained effect (cost 2 focus): `maintainedEffect` resolves to it, and the
  Maintain action appears with `cost.focus: 2`.
- Using the Maintain action: focus paid 3 → 1 via the event stream, and the effect's
  duration re-anchored to the current round.
- Survival/expiry cadence: maintained in round N → alive at round N+1 turn start (remaining
  1) → skipping the Maintain action in round N+1 → the effect is deleted at the round N+2
  turn start, its statuses clear, `maintainedEffect` resets, and the Maintain action
  disappears.
- Hands-only maintenance (grapple effects, `maintenance.hands` without a cost) is untouched.

maintain-action-available
*The Maintain action appears under Favorite Actions with its cost chips (Self · 0 Action ·
2 Focus) while the Chessman maintains an Invisibility effect. Focus pool: 3.*

maintain-action-used
*After confirming the Maintain action: the focus pool dropped 3 → 1 (the maintenance cost
paid through the event stream) and the effect's duration was re-anchored to the current
round. The focus cost chip highlights red while focus is too low to sustain again.*

- The 2-round duration is what gives the player their "one full turn to sustain" window; a
  1-round duration would expire at the start of the very next turn before the Maintain
  action could be taken. Happy to revisit the exact cadence.
- Point 5 of the issue ("prompt you to use the Maintain action at the start of your turn")
  is future work — the action is auto-favorited so it surfaces while available.
- Existing maintained effects created before this change carry indefinite durations; they
  will no longer be prompted for maintenance and should be removed manually (they are
  visible on the actor's effects list).
- Pre-existing quirk noticed while testing: effects created on unlinked tokens via the API
  are not registered in core's ActiveEffect expiry registry, so core's turn-start expiry
  pass skips them until registration. This affects all effect durations equally (including
  Defend's guarded) and is not introduced by this PR.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -71,6 +71,10 @@
         "Description": "Spend 10 minutes outside of Combat recovering from exertion to fully restore Health, Morale, Action, and Focus.",
         "Name": "Recover"
       },
+      "Maintain": {
+        "Description": "Spend the Focus cost of a maintained effect to sustain it for another round.",
+        "Name": "Maintain"
+      },
       "Reload": {
         "Description": "Reload a ranged weapon which features a reloading time.",
         "Name": "Reload Weapon"
@@ -434,6 +438,7 @@
       "AuraNoToken": "You cannot use an Aura-target action without having a token on the scene!",
       "BadStatus": "You may not perform {action} while {status}.",
       "CannotAffordCost": "{name} has insufficient {resource} to use {action}!",
+      "NoMaintainedEffect": "You have no effect which can be maintained!",
       "CannotAffordHands": {
         "zero": "{name} must have {hands} free hands to use {action}!",
         "one": "{name} must have {hands} free hand to use {action}!",
@@ -520,8 +525,8 @@
     "DurationSeconds": "{value}S",
     "DurationUntilEnded": "Until Ended",
     "EffectSpecific": "Effect: {effect}",
-    "MaintainContent": "<p>Spend {cost} Focus to maintain {effect}?</p>",
-    "MaintainTitle": "Maintain Focus: {effect}",
+    "MaintainReplaceContent": "<p>Using {action} will end your maintained {effect} at the end of this action.</p>",
+    "MaintainReplaceTitle": "Replace Maintained Effect",
     "NoTargets": "No Targets",
     "Overview": "Action Overview",
     "PlanMovement": "Plan Movement",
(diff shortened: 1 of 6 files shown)


roth-michael:
Without taking too deep a dive into the code itself, I think ideally we'd handle expiry a little differently. Two ideas come to mind, I'm not sure which is better. My instinct is that option (2) below is more in line with the intended behavior.
1. Instead of turn start, turn end expiry. This way if you don't maintain an effect on your turn, it is removed at the end of your turn.
2. Continue to prompt for maintenance at the start of your turn; saying no still removes the effect immediately.

The only difference between (2) and the current behavior is that (2) would fire off an explicit "Maintain" action workflow, and perhaps that we create a more generic mechanism for being prompted to do something, rather than having the current hardcoded "is this actor maintaining an effect, if so create this specific dialog" logic. As-is in this PR, I don't love that you can accidentally forget about maintaining, and I don't love that the choice _not_ to maintain doesn't take effect until the following round. I wouldn't necessarily get to implementing a new solution until @aaclayton weighs in, as he may disagree with my thoughts on this.

PR #1345, opened 2026-06-22
a big batch of updating descriptions

danielrab:
mainly includes adding `<strong>` tag to stuff, but also some usage of `@Condition` and `@Action` enrichers

PR #1509, opened 2026-09-14
Adversary Talent descriptions pass

roth-michael:
For this and all following "description pass" PRs, it'll just be me going through #1345 one chunk at a time, applying Daniel's suggestions unless I notice something I think should be further changed or was missed. I won't be going over any of the documents that aren't in that PR, so this won't be exhaustive, but at the end it should be a collection of PRs representing the original.


aaclayton (approved):
Great

PR #1526, opened 2026-09-17
Implement logic for behavior creation when using Actions

roth-michael:
Split out this bit of logic for more digestible review.


aaclayton (approved):
Good iteration on this. Merging, and I can apply some cleanups afterwards.

